PR #11, opened 2015-08-30
Improvements for testdrive: Firefox compatibility & JSON parsing

jnv:
I have noticed that the font preview is broken in Firefox due to incorrect `src` declaration in `@font-face`, so I have fixed it and made also a small improvement to the charmap code.

I'm not sure though if the `gh-pages` branch is generated from the `testdrive/index.html`. Should I send a new PR against that branch?

### DIFF
--- a/testdrive/index.html
+++ b/testdrive/index.html
@@ -270,8 +270,7 @@
 				var sample = $('<samp>').append('<span class="langname">' + record.lang + '</span><span class="sentence">' + record.sample + '</span>');
 				div.append(sample);
 			});
-			$.get(font.name + '.charmap', function(data){
-				data = JSON.parse(data);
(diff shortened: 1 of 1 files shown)


jnv:
By the way, in the CSS, you [declare a fallback font to `monospace`](https://github.com/be5invis/Iosevka/blob/dd0eae9ecf04e3199e8a2f48f032799a2f9a7173/testdrive/index.html#L67); this is definitely the best practice for the web, but in case of the font preview it can be misleading for user: in case the webfont is not loaded, you get system's default monospace font. Perhaps it would be better to omit the fallback font?

be5invis:
Well @jnv the reason I do not use `$.getJSON` is that someone on Hacker News reported that this procedure fails, but after I switched into `$.get` the problem is solved.

PR #32, opened 2015-11-30
Create font-iosevka.rb

chuchana:
Added formula for Homebrew/Caskroom-fonts.

For this to appear in Caskroom-fonts, you will need to create a pull request at https://github.com/caskroom/homebrew-fonts.


be5invis:
Well I do not think that it is suitable to put a package manager publishment script in my reposity. Maybe you can establish a new repo with the latest font releases and publish it into package managers.

PR #44, opened 2015-12-23
Add underscore.low variant, selectable via v-underscore-low.

bitserf:
Hello, 

Like the low variants of ~ and *, I have created a variant for underscore that puts it below the baseline.
I copied the approach used by your other variant selectors, if I am doing it incorrectly, please let me know. I put it in the same groups as the other high/low selectors.

I am not sure what `[Upright]` or `[Italify]` will do. Also, I was only guessing that `[Translate 0 ...]` means translate position around the Y axis, so please correct me if this is mistaken.

Thanks!

### DIFF
--- a/glyphs/symbol-punctuation.ptl
+++ b/glyphs/symbol-punctuation.ptl
@@ -414,11 +414,17 @@ symbol-block 'Typographic Symbols'
 
 
 symbol-block 'Dashes'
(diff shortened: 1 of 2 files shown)


be5invis:
The `[Upright]` and `[Italify]` is necessary to translate something in italic and oblique shapes. Anyway, merged and will be released in the next release.

PR #54, opened 2016-01-22
Typo in parameters file - weight spelled incorrectly

IceMan81:


be5invis:
Thanks. And it fixes the incorrect OS/2 weight class value in Medium weights.

PR #179, opened 2017-04-02
Add stylistic set ss10 for 'Envy Code R' style

damieng:
Hi, really like Iosevka!

[Envy Code R](https://damieng.com/envy-code-r) is a font I created many years ago that would work well as a stylistic set for Iosevka as it isn't that close to any of the other stylistic sets (uses hooked il)

Input Sans Mono has a [Envy Code R style](http://input.fontbureau.com/preview/?size=14&language=python&theme=solarized-dark&family=InputSans&width=300&weight=300&line-height=1.2&a=0&g=ss&i=topserif&l=topserif&zero=slash&asterisk=0&braces=0&preset=envy&customize=please) as well.


damieng:
Thanks!

PR #186, opened 2017-04-06
Added ligations for << and <<< with varied spacing

mrkgnao:
I should've made a branch for this.


be5invis:
@mrkgnao I think `>>` and `<<` can be improved by moving **both** the parts by the half amount of kerning.

mrkgnao:
@be5invis I was considering that too. I'll do that.

PR #252, opened 2018-02-22
Fix incorrect braille patterns. 

sprang:
Many of the braille glyphs in Iosevka are incorrect. This is very noticeable in terminal programs that render with braille (e.g. https://github.com/asciimoo/drawille). This is due to dots 4, 5, 6, and 7 being drawn at the wrong coordinate.

This patch takes advantage of the inherent binary pattern in the Unicode braille sequence to simplify glyph generation. I'm not sure what the cmpPriority and flatten settings are meant to do, so the code that sets those values may need further tweaking. For now, I mapped the old glyphs to the new equivalents.


be5invis:
`cmpPriority` and `flatten` is used in the "autoref" step, used to prevent deep reference nesting.
Thank you for your PR.

PR #256, opened 2018-03-02
[DO NOT MERGE] Cyrillic transliteration of Latin alphabet

lotem:
I did this just for fun.

I open this pull request to share some of my takeaways while trying to understand and modify code, as well as the resulting code to anyone who's interested.
Feel free to close the pull request when it's no more interesting.

What's done here is simply remapping Cyrillic glyphs to code points of non-strictly corresponding Latin letters. I chose Cyrillic letters because their glyphs are available in the font already. I hope next time I could create and remap new glyphs like Bopomofo symbols, which people say I'm good at handwriting.

Unfamiliar with the font-generation code, I was initially confused by glyph names and Unicode for Latin letters.
I first tried to rename the glyphs `[A-Za-z]` by adding the prefix `latin`; with the help of error messages I could identify every occurrence of glyph names and Unicode for char respectively.

To remap the letters, add/remove assigned Unicode to affected glyphs right before building them into fonts.

That's all. Sincerely hope it amuses you.


be5invis:
This is a breaking change and I have to close it.
A very interesting idea, but not suitable for Iosevka itself.
Closed.

lotem:
Of course. Thank you.

Issue #267, opened 2018-05-01
Slimming down


moosotc:
It would be great if there was an option to subtract features: i.e.
 I don't need hinting (sufficiently high DPI monitor)
 I don't need oblique variants

On my, admittedly not bleeding edge, machine building Iosevka takes a rather long while, furthermore i had to install some packages by hand because my (admittedly bleeding edge) distro didn't have them (ttfautohint is among them) so far i have only found a way to reduce the number of weights Iosevka  is built for (c.f. https://github.com/moosotc/snippets/blob/master/bin/doiosevka) it would be great if make machinery allowed suppressing undesired styles too.

be5invis:
😑  I think have to rewrite the building system

moosotc:
Belleve Invis <notifications@github.com> writes:

> 😑  I think have to rewrite the building system

That'd be wise.

-- 
mailto:moosotc@gmail.com

alexeyten:
Similar to #243

You could comment out oblique here https://github.com/be5invis/Iosevka/blob/master/maker.js#L10

And there is option to build only certain weights

moosotc:
Alexey Ten <notifications@github.com> writes:

> Similar to #243
>
> You could comment out oblique here
> https://github.com/be5invis/Iosevka/blob/master/maker.js#L10

Thanks for the pointer.

>
> And there is option to build only certain weights

Thank i know of (and use).

-- 
mailto:moosotc@gmail.com

PR #288, opened 2018-07-26
Build only needed weights and slants

alexeyten:
Allow to choose what weights and slants to build.
E.g. I build special version for terminal and only need regular and bold weights. But also I build browser version and it's not convenient to comment and uncomment weights.


alexeyten:
Should help with #267

Issue #105, opened 2016-09-02
Line height / Vertical alignment


firedev:
* Your font version: Release
* Your font variant: Any
* Your operating system (name and version): OS X 10.11.6 El Capitan
* Your application using Iosevka: Terminal, iTerm

---

Hi, thanks for free Pragmata, the only thing I wish is x-height is to be a little bit bigger...

Could you please explain line-height/vertical alignment behaviour in terminals? This is the case with other fonts as well but I hope it can be fixed with Iosevka.

Terminal:
![image](https://cloud.githubusercontent.com/assets/115612/18189703/58a6d946-70e7-11e6-80ed-78f93601367f.png)
![image](https://cloud.githubusercontent.com/assets/115612/18189805/29dc400a-70e8-11e6-8a79-030cfda2a078.png)

Iterm:
![image](https://cloud.githubusercontent.com/assets/115612/18189722/774631f8-70e7-11e6-90f5-8eeb99f13672.png)
![image](https://cloud.githubusercontent.com/assets/115612/18189817/3fed0e9c-70e8-11e6-9984-fbaef7cab14b.png)

The letters are always shifted a little bit downwards in regards to the letter space. I tried to correct it by changing line height but it's never vertically aligned perfectly.

It is something that can be fixed?

Thank you.

firedev:
It seems to be ok with bigger font-size, is it a bug in terminal implementation?

![image](https://cloud.githubusercontent.com/assets/115612/18189924/16b04458-70e9-11e6-81c0-a3c318002a1c.png)

be5invis:
Perhaps related to rounding I think.
You can try to enlarge `leading` and `descenderPad` in `parameters.toml` to increase line height.

Hyzeta:
By enlarging descenderPad to 65, the font looks good in iterm, BUT may failed in other cases (Too many spaces at the bottom of the font if enlarge descenderPad. Using the origin font is fine.). I only observe this vertical-align problem in iterm, maybe iterm's font rendering should be blamed, but other fonts do not have this vertical-align problem.

Hyzeta:
First, thanks for the great work. Iosevka is a very amazing font

* Font version: 01-iosevka-2.2.1.zip
* Application: iterm2 Build 3.2.9 on macOS 10.14.5 (18F132)
* Font size: Regular 12

[...]

Issue #80, opened 2016-05-01
Bottom part of certain letters get cut off


eizua:
In some text editors I use, the bottom part of certain letters ("g", "y", "p", etc.) get cut off because the space between each line of text is rendered too close. Here are some screenshots from editors Emacs, Vim, and Textadept:

Emacs: 
![emacs_snippet](https://cloud.githubusercontent.com/assets/19015730/14943477/1e188cc4-100d-11e6-8ca9-f78a57517624.png)

Vim:
![vim_snippet](https://cloud.githubusercontent.com/assets/19015730/14943480/34dd2712-100d-11e6-8562-5fdd1e5b350d.png)

Textadept:
![ta_snippet](https://cloud.githubusercontent.com/assets/19015730/14943484/43388dc4-100d-11e6-9ca9-1ea410240761.png)

Emacs renders the font correctly, but in Vim and Textadept, they get cut off. These are at font size 9.
If I increase it to 10 for Vim and Textadept, the problem goes away. However, the reason I use Iosevka Term is because it's thin enough to still be readable at smaller sizes - 10 is already too large for me.

These screenshots were taken from a machine running Ubuntu MATE 16.04, but I've experienced this issue even when it was still on 14.04. I can post my config files for these editors if necessary, but I'm not really doing anything special (other than the usual keybindings/plugin loading/colorscheme setting stuff).

Any thoughts on how this can be fixed (other than switching editors)? Thanks in advance!

be5invis:
Rounding error related to line height I think.
Given that Iosevka is not designed for extremely low text size, you can try to build your variant with leading increased. (In `parameters.toml`, increase the `leading` parameter.)

eizua:
Installed Node 6.0. Built FontForge, `ttfautohint` 1.5, `otfcc` 0.1.6 manually. Followed the steps provided [here](https://github.com/be5invis/Iosevka#building-from-source).
Using `make STYLE_COMMON=term DONTHINT=1` got me the following results:

Textadept:
![donthint-ta](https://cloud.githubusercontent.com/assets/19015730/14951181/a2ba44b4-1089-11e6-9f39-ce765f841691.png)

Vim:
![donthint-vim](https://cloud.githubusercontent.com/assets/19015730/14951191/b4daed7e-1089-11e6-826d-00cee20f6884.png)

Using `make STYLE_COMMON=term` with `leading` set to 1900 in `parameters.toml` got me the following:

Textadept:
[...]

Issue #43, opened 2015-12-21
The default line height is inconsistent among weights.


davmuse:
Hello,
I've noticed a difference in the default **line-height** between the styles [thin,extra-light, light, regular] and [medium, bold, heavy]. Is there a way to **normalise** these variants ? (i would prefer a mono-spaced and mono-line-height font)
Thanks

be5invis:
In 1.4.3 the metrics in all variants and all weights are exactly same (1.25em by default).
Is it rounding error?

davmuse:
This problem appears when I use the font in OSX Pages (Yosemite) ...

be5invis:
@davmuse Have you tried to uninstall and reinstall the latest version?

bitserf:
I decided to try this (latest HEAD, revision ce20e08), and I can reproduce the issue on OS X (El Capitan). It happens independently of app, it also happens in Terminal and TextEdit, not just Pages app:

It looks like extra space is at the top for some reason, see selection:

*Regular*
<img width="341" alt="regular" src="https://cloud.githubusercontent.com/assets/167841/11987494/47ba0252-aa45-11e5-902d-1854da2a03d7.png">

*Medium*
<img width="333" alt="medium" src="https://cloud.githubusercontent.com/assets/167841/11987495/56c81658-aa45-11e5-871d-0fe8d562926b.png">

The bottom looks the same amount of space.

I have no idea what could account for the difference - I compared Pragmata Regular and Bold on OS X and it has the same spacing, so there must be something different about the two weights that causes this, maybe not metrics.

be5invis:
@bitserf So OSX calculates the default line height using something other than metrics?!
I'd like to figure out why. I'll build a variant with a character with extremely high ascender and low descender to determine whether the line height is decided by the maximum bounding box of the glyphs.

be5invis:
@bitserf @davmuse I added a glyph (U+E09F) with extremely large bounding box. Is this version still suffering from line height inconsistency?
Download: http://7xpe0v.com1.z0.glb.clouddn.com/iosevka-test-lineheight-osx-20151224a.7z

bitserf:
Here's another funny thing. I wrote a test application to render fonts for comparison, and it seems Safari corrects the baseline, so it must be something in Core Text (OS X text renderer). 

Test app:
https://gist.github.com/bitserf/ec3fc3ce76b8231dfa99

Result PDF:
[...]

PR #339, opened 2019-06-26
Fix vertical alignment

Hyzeta:
Iosevka is a great font, but it has vertical alignment problem on macOS, see 

- <https://github.com/be5invis/Iosevka/issues/105>
- <https://github.com/be5invis/Iosevka/issues/43>
- <https://github.com/be5invis/Iosevka/issues/80>

It seems that Iosevka use descenderPad to fix the vertical alignment problem, by enlarging descenderPad. 

I did some experiments and measured in fontforge, by enlarging descenderPad to 65, the alignment problem will be fixed.

However, only enlarge descenderPad will result in new problem (The bottom padding will be too large on some editor). This is because that some text render will use ascent/descent/lineGap. I have checked google's roboto font and apple's SF font, these fonts's lineGap was set to zero.

With lineGap set to zero, and descenderPad set to 65, the vertical alignment will be fixed without introducing new problems.

**Before fix**

**After**

For now I only test the fixed-built on macOS and Windows 7 Virtual Machine, further check is needed.

You can download the fixed-built:
[iosevka-hyzeta.zip](https://github.com/be5invis/Iosevka/files/3329924/iosevka-hyzeta.zip)

### DIFF
--- a/meta/aesthetics.ptl
+++ b/meta/aesthetics.ptl
@@ -181,8 +181,8 @@ export : define [setFontMetrics para metrics font] : begin
 	set font.OS_2.usWinDescent      ([Math.abs desc] + descenderPad)
 	set font.OS_2.sTypoDescender    (desc - descenderPad)
 
-	set font.hhea.lineGap           (para.leading - asc + DESCENDER)
-	set font.OS_2.sTypoLineGap      (para.leading - asc + desc)
+	set font.hhea.lineGap           0
+	set font.OS_2.sTypoLineGap      0
 
 	set font.OS_2.sxHeight          XH
 	set font.OS_2.sCapHeight        CAP
--- a/parameters.toml
+++ b/parameters.toml
@@ -6,7 +6,7 @@ designer = 'Belleve Invis'
 description = 'Spatial efficient monospace font family for programming. Built from code. http://be5invis.github.io/Iosevka'
 
 leading = 1250     # Default line height times 1000.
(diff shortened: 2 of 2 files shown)


be5invis:
Have you verified that this won’t break layout on other platforms?

be5invis:
Please, test the box-drawing characters on Mac, Windows and Linux, especially on older software like Windows’ console (7 and 10).

Hyzeta:
Ok, I will test on other system, post the screenshots, make sure it won't break on other platforms.

# TO-DO LIST

## macOS

- [ ] iterm
- [ ] sublime text
- [ ] vscode
- [ ] macOS terminal

## Linux (Arch linux with gnome desktop)

- [ ] gnome terminal
- [ ] vscode

## Windows

- [ ] windows 7 console
- [ ] windows 10 console
- [ ] vscode

Anything missing?

be5invis:
These should be fine. Note that you are changing metrics so make sure you are not breaking things, esp. for box-drawing characters, since they heavily relies on the font’s internal metrics.

lshoravi:
Here's KDE's console on Manjaro, for testing.

file:///tmp/Spectacle.xdHJqC/Screenshot_20190714_113233.png

be5invis:
Image link seems broken.

lshoravi:
I messed up the link- edited now.

be5invis:
LGTM. MERGE!

Issue #81, opened 2016-05-02
Consolidation of Symbol Requests


be5invis:
| ✓ | Ch. | Unicode | Name | Proposed Glyph Name |
| --- | --- | --- | --- | --- |
| |⋈|U+22C8| | |
| |⋉|U+22C9| | |
| |⋊|U+22CA| | |
| |⸘|U+2E18|Gnaborretni| |
| |⇊|U+21CA|Downwards Paired Arrows| |
| |⤤|U+2924|North East Arrow with Hook| |
| |↧|U+21A7|Downwards Arrow from Bar| |
|  | `➦` | U+27A6 | Heavy Curved Black Upwards and Rightwards Arrow | /uptorightarrowsymb |
|  | `⚙` | U+2699 | Gear | /gear |
|  | `🌡` | U+1F321 | Thermometer | /thermometer |
|  | `⌛` | U+231B | Hourglass | /hourglass |
|  |      |    | Half brackets and integrals | |
| 1.13.3 |◐|U+25D0|Circle with Left Half Black| |
| 1.13.3 |◑|U+25D1|Circle with Right Half Black| |
| 1.13.3  |◒|U+25D2|Circle with Lower Half Black| |
| 1.13.3  |◓|U+25D3|Circle with Upper Half Black| |
| 1.13.3  |◔|U+25D4|Circle with Upper Right Quadrant Black| |
| 1.13.3  |◕|U+25D5|Circle with All But Upper Left Quadrant Black| |
| 1.11.4 | `❯`| U+276F | HEAVY RIGHT-POINTING ANGLE QUOTATION MARK ORNAMENT| (auto) |
| 1.10.2 | ` ` | Various | APL Symbols | (no name assignment) |
| 1.10.1 | `‽` | U+203D | Interrobang | /interrobang |
| 1.9.1 | `℃` | U+2103 | Degree Celsius | /degC |
| 1.9.1 | `✚` | U+271A | Heavy Greek Cross | /heavygreekcross |
| 1.9.1 | `✖` | U+2716 | Heavy Multiplication X | /heavycross |
| 1.9.1 | `⬡` | U+2B21 | White Hexagon | /whitehexagon |
| 1.9.1 | `⬢` | U+2B22 | Black Hexagon | /blackhexagon |
| 1.9.1 | `⬣` | U+2B23 | Horizontal Black Hexagon | /hblackhexagon |
| 1.8.5 | `⚑` | U+2691 | Black Flag | /blackflag |
[...]

PR #340, opened 2019-06-29
Add half-box triangle glyphs

robertgzr:
As seen here (\uE0B8-\uE0BF):
https://github.com/ryanoasis/powerline-extra-symbols#glyphs

### DIFF
--- a/glyphs/symbol-geometric.ptl
+++ b/glyphs/symbol-geometric.ptl
@@ -1381,5 +1381,63 @@ export : define [apply] : begin
 		line-to    (RIGHTSB - OX) parenBot
 		line-to    (SB + OX) parenBot
 		save 'pwlLock' 0xE0A2
-	
+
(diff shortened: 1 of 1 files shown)


robertgzr:
resolves https://github.com/be5invis/Iosevka/issues/81#issuecomment-243671896

robertgzr:
The slashes could maybe be 2 glyphs instead of 4, since they seem to be identical pairs of eb09/e0bf and e0bb/e0bd

and I first tried to base all of the triangles on the bottom-right glyph by just flipping them but that threw off the alignment of the top glyphs.

What I essentially put in there before reverting to the current code was:
```
sketch # pwlCornerTopLeft
               include : ForceUpright
               include : pwlCornerBottomLeft
               include : FlipAround MIDDLE [mix pwlNear pwlFar 0.5]
               save 'pwlCornerTopLeft' 0xE0BC
```

PR #354, opened 2019-09-20
Move 'colors' from devDependencies to dependencies

rileyinman:
Because `checkenv` is run on install and requires `colors/safe`, trying to build and install without installing dev dependencies fails. Moving it to `dependencies` solves the issue.


rileyinman:
I think I misinterpreted the distinction between `dependencies` and `devDependencies`, my bad!

PR #363, opened 2019-10-06
Nine variant

robertgzr:
Adds a variant of the number nine glyph that is the number six glyph rotated on its head:

image

*I also changed the default glyph variant of seven to use `standard` instead of `normal` to be more in line with other glyphs, see `r`.*

*ps: if you have any suggestions for a better name than `180six` I'm all for it, I just couldn't think of a better one^^*

### DIFF
--- a/glyphs/numbers.ptl
+++ b/glyphs/numbers.ptl
@@ -175,7 +175,7 @@ export : define [apply] : begin
 		reverse-last
 		branch
 			if SLAB : include : DownwardLeftSerif (SB + OXE) CAP VJUT
-			save 'seven.normal'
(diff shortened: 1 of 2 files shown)


Tazshelby2016:
What is the state of progress on this pull request?

robertgzr:
@be5invis friendly ping on this? Is this the right approach? Do you even want it in Iosevka?

be5invis:
I have my better impl. Closed "like merged".

chelovechishko:
`cv97` has wrong description.

be5invis:
@chelovechishko Fixed in latest commit.

PR #464, opened 2020-03-28
Update index.html

ilizol:
remove trailing space


be5invis:
The HTML files are generated from EJS. Closed.

PR #498, opened 2020-04-17
Add Bashkir to supported languages

AlsoScratch:
Sorry, I couldn't bother to count the languages.


be5invis:
The supported language list in README (as well as website) is automatically generated from CLDR database. They will get overwritten next time.
If you want to add override please modify `utility/export-data/supported-languages.js`.
Also, **do not raise PR to master. Raise to dev.**

AlsoScratch:
So how do I modify that particular file?

be5invis:
@AlsoScratch This is around L4:

```js
// List all the languages that Iosevka supports, but cannot inferred from CLDR data.
const overrideSupportedLanguages = [];
```

You can change like this:

```js
// List all the languages that Iosevka supports, but cannot inferred from CLDR data.
const overrideSupportedLanguages = ["Bashkir"];
```

PR #538, opened 2020-05-19
Fix typo in private build plans sample comments

akarzim:


alexeyten:
Actually it’s both. Weights and widths

PR #661, opened 2020-08-10
Fix typo in README.md

alexeyten:


be5invis:
Please modify `params/variants.toml` instead, since that is the true source of these variants' description.

alexeyten:
Fixed `params/variants.toml` too

PR #749, opened 2020-11-21
Add more metric overrides

alexeyten:
Allow override of `SymbolMid` and `parenSize`.


be5invis:
Interesting, `SymbolMid` is uppercase. Made a change to turn it into lowercase.

PR #750, opened 2020-11-21
Allow inherit ligations from `calt`

alexeyten:
Add `calt` to `hives` so `inherit = 'calt'` works.

Also get rid of `optInBuildup`/`optOutBuildup` because they just merged into `para.ligation.defaultBuildup`.

### DIFF
--- a/font-src/support/ligation-data.js
+++ b/font-src/support/ligation-data.js
@@ -21,7 +21,7 @@ module.exports = function applyLigationData(data, para, argv) {
 	}
 
 	para.ligation = {
-		defaultBuildup: defaultBuildup,
+		defaultBuildup,
 		caltBuildup: []
 	};
 	if (argv.ligations) {
--- a/font-src/support/ligation-data.js
+++ b/font-src/support/ligation-data.js
@@ -3,8 +3,7 @@
 const Parameters = require("./parameters");
 
 module.exports = function applyLigationData(data, para, argv) {
-	const optInBuildup = {};
-	const optOutBuildup = {};
+	const defaultBuildup = {};
 
 	const hives = {};
 	hives["default"] = { caltBuildup: [] };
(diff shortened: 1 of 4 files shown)


be5invis:
Well in ligation settings you are *creating* a "calt" configuration.
It would be better to inherit "calt-default" instead.

be5invis:
Also I think it may worth to rename these internal "hives"' prefixes, like into `ligset-inherit-<>`, `ligset-enable-<>` and `ligset-disable-<>`.

font-src/support/ligation-data.js
```diff
@@ -17,18 +16,12 @@ module.exports = function applyLigationData(data, para, argv) {
...
-		if (!comp.isOptOut) {
-			hives["ligset-" + gr] = { caltBuildup: ligSets };
-		}
+		defaultBuildup[comp.tag] = ligSets;
+		hives["ligset-" + gr] = { caltBuildup: ligSets };
 	}
 
 	para.ligation = {
-		defaultBuildup: { ...optInBuildup, ...optOutBuildup },
+		defaultBuildup: defaultBuildup,
```
be5invis, line 24: No need to duplicate this variable name. Use `{ defaultBuildup }`.

font-src/support/ligation-data.js
```diff
@@ -17,18 +16,12 @@ module.exports = function applyLigationData(data, para, argv) {
...
-		if (comp.isOptOut) {
-			optOutBuildup[comp.tag] = ligSets;
-		} else {
-			optInBuildup[comp.tag] = ligSets;
-		}
-		if (!comp.isOptOut) {
-			hives["ligset-" + gr] = { caltBuildup: ligSets };
-		}
+		defaultBuildup[comp.tag] = ligSets;
+		hives["ligset-" + gr] = { caltBuildup: ligSets };
```
be5invis, line 20: We may need to rename these internal hive names, like into `ligset-inherit-<>`, `ligset-enable-<>` and `ligset-disable-<>`.

alexeyten:
> Well in ligation settings you are _creating_ a "calt" configuration.
> It would be better to inherit something like "calt-original" or "calt-built-in" instead.

I'm not sure how to do it right.

Maybe we should inherit `calt` by default and have something like `inherit: 'none'` to clean it up.

be5invis:
@alexeyten Inheritance should be opt-in.
An interesting solution is that to rename `[composite.calt]` into `[composite.calt-builtin]` in `params/ligation-set.toml`. OT tags are stored as properties so no worries about tag compatibility.

alexeyten:
Maybe `[composite.default]` since it is the default.

be5invis:
@alexeyten `default-calt` seems better.

PR #778, opened 2020-12-22
Add raised-turn-comma variant for ascii-grave

agspathis:
This glyph is commonly used in Lisp literature (e.g. PAIP) for backquote.

### DIFF
--- a/params/variants.toml
+++ b/params/variants.toml
@@ -1798,6 +1798,11 @@ rank = 2
 description = "Show ASCII grave (`` ` ``) as raised comma."
 selector.asciiGrave = 'raisedInverseComma'
 
+[prime.ascii-grave.variants.raised-turn-comma]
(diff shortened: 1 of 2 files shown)


params/variants.toml
```diff
@@ -1798,6 +1798,11 @@ rank = 2
 description = "Show ASCII grave (`` ` ``) as raised comma."
 selector.asciiGrave = 'raisedInverseComma'
 
+[prime.ascii-grave.variants.raised-turn-comma]
+rank = 3
+description = "Show ASCII grave (`` ` ``) as raised turned comma."
```
be5invis, line 1803: May worth noting “identical to curly open single quote symbols (U+2018)”

agspathis: It's a useful hint, indeed, added!

PR #1282, opened 2022-01-25
Fix reference to D (cv04) in variants.toml

protesilaos:


be5invis:
Please submit a PR to `dev` instead.

PR #1283, opened 2022-01-26
Fix reference to D (cv04) in variants.toml

protesilaos:


protesilaos:
This is the same as #1282 but for the `dev` branch.

PR #1328, opened 2022-03-20
Add notice about concurrently running jobs

leduyquang753:
This pull request adds a notice to the build instructions page about the possibility of saturating system resources of the Verda build system and how to resolve the issue. The command-line argument seems to be undocumented making it very valuable for people with constrained system resources.


be5invis:
Please submit to `dev` instead.

PR #1387, opened 2022-07-23
Fix assignment of `l` for SS08

hwadii:
Since v0.829 of Pragmata Pro, `l` looks more like `zshaped` than `serifed`. See [here](https://fsd.it/pragmatapro/Handbook.png).


be5invis:
Please submit to `dev` instead.

Issue #1122, opened 2021-06-15
Reduce repo size


MithicSpirit:
**Is your feature request related to a problem? Please describe.**
The current repository is 17 GB, the vast majority of which is in the `.git` folder. This means that it takes around half an hour to clone the repo for just 9 MB of data (as well as effectively wasting a lot of disk space). I am aware that it is possible to download a .zip through github that does not include the `.git` folder, but this makes updating to a different commit quite annoying.

**Describe the solution you'd like**
Rewriting the history could greatly reduce the repository size. This can be done with `git filter-branch` (included in `git` by default)  or `git filter-repo` (requires separate installation but recommended).

**Describe alternatives you've considered**
There might be other ways to reduce repo size but I'm not aware of them.

be5invis:
Use [Git clone with depth limit](https://git-scm.com/docs/git-clone#Documentation/git-clone.txt---depthltdepthgt).

OnkelTem:
32Gb now. It's crazy.

PR #1395, opened 2022-08-11
Add note to build docs on limiting cloned size via --depth

natejlong:
When cloning, I had assumed the repo's files were just huge for some reason until I saw in https://github.com/be5invis/Iosevka/issues/1122 that it was really just the commit history

This publicizes that more and may save others some bandwidth.


be5invis:
Please submit PR to `dev` instead

Issue #1326, opened 2022-03-08
Issue with ligatures on kitty


tpoisot:
After using a private build plan, the ligatures do not work in kitty:

![1646763923](https://user-images.githubusercontent.com/579329/157316169-df4991bf-4920-40e3-aa87-678aa193e0ba.png)

---

* Your font version: 721fe3367f8a9e0a9107463584ba22c0dce27961
* Your font variant: Custom
* Your operating system (name and version): Linux 5.16.12-arch1-1
* Your application using Iosevka: `kitty 0.24.4`

If you have problem when building Iosevka, please provide these information:

* Your Node.js version: `v17.6.0`

* Your `private-build-plans.toml`, if present:

~~~
[buildPlans.iosevka-custom]
family = "Iosevka Custom"
spacing = "normal"
serifs = "sans"
no-cv-ss = true

  [buildPlans.iosevka-custom.variants]
  inherits = "ss08"

    [buildPlans.iosevka-custom.variants.design]
    a = "double-storey-serifless"
    d = "toothed-serifless"
    f = "tailed"
    g = "single-storey-flat-hook-serifless"
    h = "straight-motion-serifed"
    i = "serifed-flat-tailed"
    l = "serifed-flat-tailed"
    p = "eared"
    q = "straight"
    t = "flat-hook"
    v = "curly"
    x = "curly-serifless"
    y = "curly-turn"
    z = "straight-top-serifed"
    turn-v = "curly"
    lower-delta = "flat-top"
    capital-gamma = "top-right-serifed"
    lower-lambda = "curly-turn"
    lower-xi = "rounded"
    zero = "slashed-split"
    one = "base"
    two = "curly-neck"
    four = "semi-open"
    five = "oblique-upper-left-bar"
    seven = "curly-serifed"
    eight = "crossing-asymmetric"
    asterisk = "turn-hex-low"
    underscore = "high"
    pilcrow = "high"
    caret = "low"
    paren = "flat-arc"
    brace = "curly"
    number-sign = "upright"
    ampersand = "et-toothless-corner"
    at = "fourfold-solid-inner"
    dollar = "open-cap"
    cent = "bar-interrupted"
    percent = "rings-continuous-slash"
    lig-ltgteq = "slanted"
    question = "corner-flat-hooked"

[buildPlans.iosevka-custom.weights.light]
shape = 300
menu = 300
css = 300

[buildPlans.iosevka-custom.weights.regular]
shape = 400
menu = 400
css = 400

[...]

PR #1402, opened 2022-08-21
Add export-glyph-names to sample.toml

real-or-random:
and also add a reference to the full docs

Consistent with https://github.com/be5invis/Iosevka/issues/1326#issuecomment-1221396681


be5invis:
Please submit PRs to `dev` branch instead.

PR #1558, opened 2023-02-19
fix Partial Differential glyph

### LINKED ISSUE #1503: Partial Differential glyph shape, again (U+2202)
I recognize that the current shape of the partial differential symbol is [due to this issue](https://github.com/be5invis/Iosevka/issues/862). This issue might be in direct opposition to that suggestion (the old one definitely makes more sense).

This is what the [Wikipedia page](https://en.wikipedia.org/wiki/%E2%88%82) has to say for the character:
> The character ∂ (Unicode: U+2202) is a stylized **cursive** d mainly used as a mathematical symbol

> The symbol is variously referred to as "partial", **"curly d", "rounded d", "curved d"**...

The current design looks nowhere "cursive" nor "rounded". I'd argue it's even unrecognizable as the partial d symbol.

The shape change seems to take the phrase "closer to a mirrored 6" a bit too literally: It's now like a mirror of the default straight-bar style 6 (I can't tell if it is *completely* straight), which looks really weird since literally no fonts I found renders the character like that, even those provided by the original issue.

Since no more variant selectors can be added (I don't think binding them to the variants of 6 is a good idea either), I'd suggest at least making the ascender bend and look more like ð (or italic *д*) instead, as mentioned in that issue.

Or just revert to the old shape, idk.

jmcwilliams403:
fixes #1503

### DIFF
--- a/font-src/glyphs/symbol/letter.ptl
+++ b/font-src/glyphs/symbol/letter.ptl
@@ -173,29 +173,7 @@ glyph-block Symbol-Letter : begin
 		include : HSerif.rb (Middle + sw / 2 * HVContrast) 0  (LongJut / 2)
 		include : DotAt Middle (XH + AccentStackOffset) (DotRadius * sw / Stroke)
 
-	create-glyph 'partial' 0x2202 : glyph-proc
-		include : MarkSet.b
-		local fine : Stroke * CThin
-		local coFine : mix fine Stroke 0.5
-		local yMid : Ascender * 0.6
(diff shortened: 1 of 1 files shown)


be5invis:
I suggest to not delete the old glyph -- instead, creating a variant for `partial`

jmcwilliams403:
> I suggest to not delete the old glyph -- instead, creating a variant for `partial`

I wouldn't know how to do that. Do you think it's possible that you could?

be5invis:
@jmcwilliams403 Then do not delete the old code, just comment it out, and I will do the rest.

Issue #1461, opened 2022-12-12
Cyrillic Extended D


Logo121:
- The requested character is...
  - [ ] Latin
  - [x] Cyrillic
  - [ ] Greek
  - [ ] Greek
  - [ ] Punctuation
  - [ ] Symbol
- [ ] Some other monospace/programming fonts supported this character. Provide images below.

------

no justification, honestly.

But given how you seem to have no trouble supporting almost the entire Latin Extended F, I guess (most of) [this block](https://www.unicode.org/charts/PDF/U1E030.pdf) is not too difficult to support.

It is a block added in the latest version (15.0), consisting of _only_ superscripts and subscripts (and one combining character), of characters mostly already supported by this font.

The only exceptions are the 2 character based on `ꚉ` and `ꙑ`, the former of which might be a trouble. i don't know.
 
(btw, Latin Ext. F and Ext. G have hyphens in the [specimens page](https://typeof.net/Iosevka/specimen) but other extended blocks don't?)

Here is a table(?) of the corresponding base characters for each character in the block:
```
абвгдежзиклмопрс
туфхцчшыэюꚉәіјөү
ӏабвгдежзиклопсу
фхцчшъыґіѕџҫꙑұ
```

(and `і` at the end)

Also related are some superscripts (and combining characters) in Cyrillic Ext. B. All of the base characters are already supported.
![image](https://user-images.githubusercontent.com/21302803/207107503-6787fddf-dd70-467d-bf22-f3ca94bf760a.png)

be5invis:
Well Latin Ext-F are just superscripts of existing letters. Cyrl Ext. D is more complicated.

PR #1564, opened 2023-02-23
Add Cyrillic Extended-D

jmcwilliams403:
except for U+1E04A and U+1E06C
image


Logo121:
(for documentation) This addresses #1461 up to the 2 "execeptions"

Logo121:
...oh actually it looks like U+1E049 (superscript yu) is also missing.

jmcwilliams403:
> ...oh actually it looks like U+1E049 (superscript yu) is also missing.

be5invis noticed already and added that in a later commit

Issue #1597, opened 2023-03-11
A collection of characters derivable from existing characters under rigid motion(?)


Logo121:
- The requested character is...
  - [x] Latin
  - [x] Cyrillic
  - [ ] Greek
  - [x] Punctuation
  - [x] Symbol
- Some other monospace/programming fonts supported this character (eh they probably do)

------

Here is a list of characters I could find that can be probably derived by performing some non-distorting transformation on existing characters.

### Reversed (Horizontally flipped)

| Status | Derived | Base |
| --- | --- | --- |
| ✓ | `⦣` U+29A3 REVERSED ANGLE | `∠` U+2220 ANGLE |
| ✓ | `⦰` U+29B0 REVERSED EMPTY SET | `∅` U+2205 EMPTY SET |
| ✓ | `⫮` U+2AEE DOES NOT DIVIDE WITH REVERSED NEGATION SLASH | `∤` U+2224 DOES NOT DIVIDE |
| ✓ | `⯾` U+2BFE REVERSED RIGHT ANGLE | `∟` U+221F RIGHT ANGLE |
| ✓ | `Ꙅ` U+A644 CYRILLIC CAPITAL LETTER REVERSED DZE | `Ѕ` U+0405 CYRILLIC CAPITAL LETTER DZE |
| ✓ | `ꙅ` U+A645 CYRILLIC SMALL LETTER REVERSED DZE | `ѕ` U+0455 CYRILLIC SMALL LETTER DZE |
| ✓ | `Ꙕ` U+A654 CYRILLIC CAPITAL LETTER REVERSED YU | `Ю` U+042E CYRILLIC CAPITAL LETTER YU |
| ✓ | `ꙕ` U+A655 CYRILLIC SMALL LETTER REVERSED YU | `ю` U+044E CYRILLIC SMALL LETTER YU |
| ✓ | `Ꙡ` U+A660 CYRILLIC CAPITAL LETTER REVERSED TSE | `Ц` U+0426 CYRILLIC CAPITAL LETTER TSE |
| ✓ | `ꙡ` U+A661 CYRILLIC SMALL LETTER REVERSED TSE | `ц` U+0446 CYRILLIC SMALL LETTER TSE |
| ✓ | `ꟻ` U+A7FB LATIN EPIGRAPHIC LETTER REVERSED F | `F` U+0046 LATIN CAPITAL LETTER F |
[...]

PR #1607, opened 2023-03-14
Two turned characters

Logo121:
Implements 2 characters i missed (U+1DF02, U+1DF10) and later added in #1597.
image


be5invis:
Please add your change note into `changes/<next-vesion>.md`.

Logo121:
okey.

Btw I assume the next version should be 21.1.0 and not 20.1.0? I changed that just in case.

be5invis:
Yes 21.1.0.

PR #1625, opened 2023-03-22
Correct glyph for U225B, re-add old form at U2A6E

jmcwilliams403:
The star-equals operator is intended to display as an equals with a literal star operator as found elsewhere in the same Mathematical Operators block. The asterisk-equals operator exists in the Supplemental Mathematical Operators block which uses the original glyph from the font.

### DIFF
--- a/font-src/glyphs/auto-build/composite.ptl
+++ b/font-src/glyphs/auto-build/composite.ptl
@@ -1053,8 +1053,9 @@ glyph-block AutoBuild-Accented-Equal : begin
 	createAccentedOp 'equal' 7 0.5 0 (aboveMarkBot - (SymbolMid - XH / 2)) : list
 		list 0x2259 {"triangularWedge.NWID"}
 		list 0x225a {"triangularVee.NWID"}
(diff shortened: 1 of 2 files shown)


font-src/glyphs/auto-build/composite.ptl
```diff
@@ -1053,8 +1053,9 @@ glyph-block AutoBuild-Accented-Equal : begin
 	createAccentedOp 'equal' 7 0.5 0 (aboveMarkBot - (SymbolMid - XH / 2)) : list
 		list 0x2259 {"triangularWedge.NWID"}
 		list 0x225a {"triangularVee.NWID"}
-		list 0x225b {"asterisk.pentaSMid"}
+		list 0x225b {"blackStar.NWID"}
 		list 0x225c {"whiteTriangleUp.NWID"}
+		list 0x2a6e {"asterisk.pentaSMid"}
```
be5invis, line 1058: Could use `asterisk.hexSMid` here for better distinction.

be5invis (changes requested):

be5invis (approved):

PR #1649, opened 2023-04-04
five ligatures

Logo121:
image
Implemented the ligatures in the Enclosed CJK Letters And Months block.


be5invis:
Thank you for your contribution but I plan to revert this.
See comment in https://github.com/be5invis/Iosevka/pull/1651.

Issue #1650, opened 2023-04-04
Square ligatures from CJK Compatibility Chararters


jmcwilliams403:
- The requested character is...
  - [x] Latin
  - [ ] Cyrillic
  - [ ] Greek
  - [ ] Punctuation
  - [x] Symbol
- [x] Some other monospace/programming fonts supported this character. Provide images below.
------
A continuation of #1649 
U+3371..U+337A, U+3380..U+33DF, U+33FF
Julia Mono:
![image](https://user-images.githubusercontent.com/37010132/229853391-6509697b-92e5-477c-9fc1-8527b09f70e8.png)

be5invis:
See comment in #1651.
Resolve as out of scope.

PR #1651, opened 2023-04-04
Some more ligatures in CJK Compatibility

Logo121:
image

Implemented all ligatures in the CJK Compatibility block that can be made directly (no `/` or powers)

Except Square Co, not sure if that (or `am` `pm` for that matter) should include the `.` and/or have it handled in a special way. Since some fonts don't contain periods for am/pm but most does for `Co.`, I just left it there.


Logo121:
umm

solves part of #1650.

Logo121:
Should it? The reference glyph does that, but quite a few fonts (including JuliaMono which you showed in the issue) doesn't.

I couldn't find any sources determining whether they should be merged or not either, so I just did it the unfused way.

jmcwilliams403:
I suppose you're right then, considering my own example doesn't even do it.

be5invis:
I plan to reject this (and revert @Logo121's previous PR) as **out of scope**.

These characters, as far as I know, are included to provide compatibility to certain early Japanese character sets (or even earlier -- Japanese phototypesetting machines). Since Iosevka's goal is NOT a CJK typeface, I'd like to consider this as outside Iosevka's scope.

For U+3251...325F and U+32B1...32BF, I plan to remove these in the official builds too. Including these characters made Iosevka being considered as a "CJK font", which leads a lot of troubles.

Logo121:
I understand your decision on the compat ligatures.

However, I have questions about the implication of removing the circled numbers in the Enclosed CJK Letters and Months block.

According to Unicode documents, the circled 21-50 are added along with the negative and double circled numbers in the Enclosed Alphanumerics block, but assigned in Enclosed CJK Letters due to the former block being fully allocated.  
It would arguably be counted as Enclosed Alphanumerics instead, if the Enclosed Alphanumerics Supplement block was allocated back then.

If "being encoded as CJK codepage compatibility" is the rationale of removal, would that imply U+24EB..24FE (negative circled 11-20, double circled 1-10) should be removed as well, along with the removal/exclusion of probably many other codepoints (in Dingbats, Enclosed Alphanumerics Supplement, Miscellaneous Technical, Emojis etc)?

If "being in a block with CJK in its name" is the rationale, wouldn't that be more of an external issue, one similar to the monospace font problem?

be5invis:
@Logo121 Well, whether to remove those is not decided yet. The main problem is that these characters are placed into the CJK BLOCK, so a lot of tools (like Google's FontBakery) will treat fonts supporting them as CJK fonts. For U+24EB..24FE they are outside the CJK block so there's no problem.

PR #1685, opened 2023-04-22
Komi Nje, Komi Lje

Logo121:
Based on Hwair (left).
image

How tall the upward hook should be is not really defined, and is different across different fonts. 
I just left it at CAP/2 or XH/2.


jmcwilliams403:
Are you sure the top serifs on the El are correct?

Logo121:
what's wrong with the serifs? It's based on other El shapes.

jmcwilliams403:
Nevermind, the specimen wasn't showing the serifs as variants unless I completely switched over to slab. Ignore what I said.

PR #1687, opened 2023-04-23
Komi Zje, Komi Dzje

Logo121:
image

Modified CyrlZe functions to accept left right and stroke parameters,  
and also a lower-ze shape that stops halfway so that the new parts (Straight line for Dzje, Upwards hook for Zje) can be stitched onto it. Not sure if this is done properly however, since idk how those knots work.

Also modified the upper shape a little bit due to some (hopefully unintended) stroke width problems with Zje.


font-src/glyphs/letter/greek/lower-epsilon.ptl
```diff
@@ -234,6 +307,6 @@ glyph-block Letter-Greek-Lower-Epsilon : begin
 		local dimUpper : CyrZeShapeDim         CAP            midGap blend
 		local dimLower : SmallEpsilonShapeDim  (CAP - midGap) 0      blend
 		include : union
-			CyrZeUpperShape        SLAB-NONE  CAP            midGap blend Hook
-			SmallEpsilonLowerShape SLAB-NONE (CAP - midGap) 0      blend Hook
+			CyrZeUpperShape        SLAB-NONE  CAP           midGap SB RightSB blend Hook
+			SmallEpsilonLowerShape SLAB-NONE (CAP - midGap) 0      SB RightSB blend Hook
```
be5invis, line 311: This corrupts U+1D24 actually.
No actions for you. I fixed it by myself in later commits.

PR #1698, opened 2023-05-01
Djerv, Stirrup r, Baseline esh

Logo121:
image

Djerv and Stirrup R uses a shared shape, based on a flipped UShape. The stroke is borrowed from the half-soft sign yer.

Though Stirrup R is a bit of a mystery.  
The only source in the Unicode proposal is sanserif, so there are no known serif version of it (no serif fonts give the shape any serif).  
The shape is not explicitly unified with Djerv, and some fonts render it *very slightly* differently from Djerv, but it's probably fine to ignore the difference.

Baseline Esh is just Esh but baseline.


jmcwilliams403:
What about combining djerv in Cyrillic Extended-A?

Logo121:
> [“Die Digitalisierung der Bestände des Tiroler Dialektarchivs: Ein Werkstattbericht zum vorläufigen Projektende”](https://www.researchgate.net/profile/Yvonne-Kathrein/publication/334888881_Die_Digitalisierung_der_Bestande_des_Tiroler_Dialektarchivs_Ein_Werkstattbericht_zum_vorlaufigen_Projektende/links/5d77f9bb92851cacdb30c65b/Die-Digitalisierung-der-Bestaende-des-Tiroler-Dialektarchivs-Ein-Werkstattbericht-zum-vorlaeufigen-Projektende.pdf) p. 65 has an example of ⟨ꭅ⟩ in a serif font. I don’t know where that font is from or how reliable it is.

This form actually looks more sensible as a "r variant", but it does look quite different from the Unicode glyph. I wonder if it's Unicode glyph that's wrong in that case? (like the multiocular O)

All fonts that support the character just follow the Unicode form (no serif in Noto Serif):
image

But idk then. Maybe be5invis can decide whether to keep that one or not.

PR #1715, opened 2023-05-11
Reversed Tse + A few decorated O's

Logo121:
image

* Minor fixes to some glyph names
* Since Tse and Reverse Tse only differ by the descender, the function to implement Rev Tse is incorporated into the existing Tse shape function
* The width of Crossed O's cross scales from 1 OverlayStroke to 0.5 OverlayStroke, so that the inner part looks less black on heavy weights.
* Unified the inner dot of Dotted Lunate Sigma, Bilabial Click and the newly implemented Monocular O.
* Tried to extend the dot function to Binocular O's but the dots become far too thin in heavier weights. The final result above already has the spacing around dots shrunk (down to 1/3 of the dot half-width), but idk if this is even legible on smaller sizes:
(Version without shrinking above vs with shrinking below, ignore the serif differences)
image
Not sure if I should make a separate thinner O glyph for each of them, whether that would affect QP versions, or if that would even make any difference in the first place.
* I'm **definitely** not going to touch multiocular O.


Logo121:
Reversed tse does not have a special italic form. At least, it's not shown to have any in the wiki or in any fonts I checked.

It's only used to transcribe a letter in a long extinct Slavic dialect that was never written in cursive in the first place.

be5invis:
> I'm definitely not going to touch multiocular O.

Well I just added that.
image

Issue #1489, opened 2022-12-20
Serif forms of some existing characters


Logo121:
- [x] The requested variant shape does not go too far away from Iosevka's design.
- [x] The requested variant does not conflict with any characters in Unicode that Iosevka currently supports.
- [x] At least two monospace/programming fonts, created by different designers, supported the requested variant. Provide images below.
(only some of them)
------

Below is a comparison between several serif (not necessarily monospace) fonts and Iosevka, on some glyphs that should have serifs, but is not given one in Iosevka. (unsupported characters are rendered in unifont instead)
![image](https://user-images.githubusercontent.com/21302803/208656382-9b7f19cf-6cf9-461c-8e47-e0831de24ab2.png)
Text: `ǷƿȜȝᵹꝽѠѡѼѽѾѿꙌꙍ`

Some notes:
- Cyrillic omega can be rendered either as that weird symmetric G, or as something like W. Both forms are correct, but I don't think you can add more variants anyway.
- The one glyph not supported by Iosevka here is [Capital Insular G](https://github.com/be5invis/Iosevka/issues/1483), for if you are going to support that as well.

be5invis:
I ususally use Brill as reference. Here are the Crillic Omegas:

![image](https://user-images.githubusercontent.com/240091/208663687-190d788a-3fc8-4562-864c-b7f2a0a8295b.png)

I don't think inward hooks in the "broad" shape are serifs, so for these I will keep the current shape.
For non-broad Omega I will adopt the W-like shape for both the cases.

Logo121:
Ok, but I don't feel like the current glyphs look like they have "inward hooks" either. 

Here's a comparison using the above fonts (+ Brill and Noto Sans) for reference (the ones on the left are the tangentially related latin omegas (Ꞷꞷ)):
![image](https://user-images.githubusercontent.com/21302803/208673384-f4dfa942-af2f-4e10-85c3-55a3404ef9f2.png)

Most fonts seem to at least make the non-greek omegas have more pronounced hooks to differentiate(?) them from the greek one.

Also these examples in the Russian Wikipedia:
![image](https://user-images.githubusercontent.com/21302803/208674412-ed8260de-2fd2-4d77-8290-1ccfeb706b82.png)

pv4:
> Here are the Crillic Omegas:

@be5invis thank you very much! I will use one of them as my profile picture :D

Issue #1517, opened 2023-01-10
Chruch Slavonic (Cyrillic) Letters


Logo121:
- The requested character is...
  - [ ] Latin
  - [x] Cyrillic
  - [ ] Greek
  - [ ] Punctuation
  - [ ] Symbol
- [x] Some other monospace/programming fonts supported this character. Provide images below.

------

(This is for backlogging mostly.) Mentioned in #1515, Hyperglot notes 4 letters necessary to support the old Cyrillic Alphabet used for Church Slavonic. This issue is a collection of these letters, + two that are related to those letters.

Monospace fonts that support them (+non-monospace for reference):
![image](https://user-images.githubusercontent.com/21302803/211622386-a6f4c0e8-72f7-4d2f-bf7e-6c40f840a57b.png)

Text: `Ꙁꙁ Ꙗꙗ Ꙃꙃ Ꙋꙋᲈᲂу`

Notes:
* Most fonts render the letter Zemlya with a long tail, but Brill renders it with a short one (resembling greek Zeta). This version is probably still correct as per wiki: ![](https://upload.wikimedia.org/wikipedia/commons/a/ae/Early_Cyrillic_letter_Zemlia.png)
* The most problematic one is probably the monograph Uk:
  * The uppercase `Ꙋ`, which may or may not have the tail (going through the O). 
  * The lowercase `ꙋ` has the two letters blending (looking somewhat like gamma), with a non-blended version `ᲈ` for contrast. So the JuliaMono's version of the former might be wrong.
  * Unicode additionally provides a "narrow o" `ᲂ` for composing the letter as `ᲂу` (like shown in the screenshot). The "narrow" part seems to be ignored by Noto Sans Mono.

cormullion:
> * So the JuliaMono's version of the former might be wrong.

Please open an issue [here](https://github.com/cormullion/juliamono) as well. 😂

be5invis:
Far-future target

Logo121:
> > * So the JuliaMono's version of the former might be wrong.
> 
> Please open an issue [here](https://github.com/cormullion/juliamono) as well. :joy:

Done.

> Far-future target

Fair enough. This is a subset of #1142 which is marked far future as well anyway.

I'll just throw in [the original proposal](https://www.unicode.org/wg2/docs/n4607.pdf) (also mentioned in the JuliaMono issue) and [this technical note for church slavonic](https://www.unicode.org/notes/tn41/tn41-1.pdf) for future reference.

[...]

Issue #1142, opened 2021-06-28
Request for Old Cyrillic letters in Cyrillic Extended-B


KSundstrom:
As a linguist sometimes working with various older texts, e.g. in Old Slavonic, I would very humbly request the more common Old Cyrillic letters, forming the first part of the Cyrillic Extended-B block, i.e. A640–A667 (the rest of the block is of course not completely irrelevant neither). The relevant Unicode chart is available at https://www.unicode.org/charts/PDF/UA640.pdf

P.S. Thank you for an excellent font I use on a daily basis!

PR #1719, opened 2023-05-12
Neutral Yer

Logo121:
image

It is described vaguely in wiki as a "Yer but with hook", used when a transcriber cannot tell whether a letter is supposed to be front or back.

This shape references that of Noto Sans Mono (same as Everson Mono and JuliaMono):
image
Which I interpret as "having a hook that extends like back yer" as an "in-between" shape of Front and Back yer.

Other renditions include a deeper, down-pointing hook (e.g. Unicode glyph, Noto Serif) and in the extreme case, a "question mark hook" by Code2000: image

Though based on the conversation back in #1489, I suppose such hooks are probably ignorable and the same hook for sansserif can be used.

Italics follow other Yer shapes, if this is important: image

(the Upper-A thing is just a fix for a closed small yus problem I discovered while looking for the other extended/QP glyph issues)


Logo121:
Btw, after this (or otherwise), we are 1 (excluding case and variants) letter away from closing #1517 (Uk), and two letters away from closing #1142 (Uk, Yn).

PR #1723, opened 2023-05-14
Add Insular D

### LINKED ISSUE #1688: Insular D (`U+A779` & `U+A77A`)
  - [x] Latin
  - [ ] Cyrillic
  - [ ] Greek
  - [ ] Punctuation
  - [ ] Symbol

------
The combining version (`U+1DD8`) is already supported by the font using a homoglyph from Cyrillic.
below is a rendering under Code 2000 with lowercase insular D followed by lowercase eth:
image
Here is the capital form compared to capital O under Noto Sans mono:
image

jmcwilliams403:
Resolves #1688

Regular:
image
Thin:
image
Heavy:
image


Logo121:
As mentioned in the issue, I would really prefer that the insular D shape not be identified with italic de, since it (the ascender) is never really depicted that way in the original script or the Unicode standard.

jmcwilliams403:
> As mentioned in the issue, I would really prefer that the insular D shape not be identified with italic de, since it (the ascender) is never really depicted that way in the original script or the Unicode standard.

It's being identified with icelandic eth, which has slightly different metrics from cyrillic de. Eth is also found in the same scripts as insular D.

Logo121:
I stand corrected, but my point is the same: The shape of eth and insular D is also differentiated beyond the stroke as [shown in the proposal](https://www.unicode.org/wg2/docs/n3122.pdf)

image
image

PR #1730, opened 2023-05-17
add, fix

jmcwilliams403:
Plain text for lookup/reference:
```
∟⯾⦢∠⦣⟋⟍
⦰∅⦱⦲⦳⦴
|‖∣∥∤⫮∦
⸶⸷†‡⸸⹋
```

Regular:
image
Italic (cv99 affected characters circled):
image
Heavy:
image
Thin:
image


nsemrau:
In the example pictures, 'EMPTY SET WITH SMALL CIRCLE ABOVE' (U+29B2) seems to be slightly cut on top of the ring in weights Regular, IItalic and Heavy. Thin seems to be okay.

jmcwilliams403:
> In the example pictures, 'EMPTY SET WITH SMALL CIRCLE ABOVE' (U+29B2) are slightly cut on top of the ring in weights Regular, IItalic and Heavy. Thin seems to be okay.

That's just microsoft notepad not being able to display the full height. I'm just using notepad for presentation.

Issue #1732, opened 2023-05-18
“Old Polish o” aka “o rogate” character request


silmeth:
- The requested character is...
  - [x] Latin
  - [ ] Cyrillic
  - [ ] Greek
  - [ ] Punctuation
  - [ ] Symbol
- [ ] Some other monospace/programming fonts supported this character. Provide images below.

------

A character used in editing Old Polish texts representing a nasal vowel in medieval Polish manuscripts. Unicode characters:
* **Ꟁ** `0xA7C0 LATIN CAPITAL LETTER OLD POLISH O`,
* **ꟁ** `0xA7C1 LATIN SMALL LETTER OLD POLISH O`,

introduced in Unicode 14 in 2021.

It’s still difficult to find any monospace font with these characters, beside Unifont.

A comparison of its shape in some other fonts (Junicode 2 Beta has a few variants for the lowercase glyph):
![](https://github.com/be5invis/Iosevka/assets/5329644/d6bedfbe-2d43-407e-86a6-cced6705357a)

Logo121:
fwiw, it's also supported by Noto Sans Mono, if you need a supporting font right now.

silmeth:
Oh, thanks! I had no idea it has the glyphs – the version of Noto Sans Mono installed with my system (2.006) is fairly old and so it didn’t show up when I was looking for fonts with it.

PR #1733, opened 2023-05-18
Polish O (#1732)

Logo121:
image

I'm not very sure if the MarkSet or the stroke lengths are correct, though it just follows Ukrainian Ge and Abkhazian Che with Descender for now

No variants are implemented. I don't think the variants implemented in Junicode are officially stated in Unicode anyway.


jmcwilliams403:
The descender/ascender of the Polish O seems to gain more thickness on heavier weights than the other e-based (es-based) glyphs.

Issue #1678, opened 2023-04-19
Novel zero variant: Zero with broken vertical middle bar


Lieutenant-L-T-Smash:
- [X] The requested variant shape does not go too far away from Iosevka's design.
- [X] The requested variant does not conflict with any characters in Unicode that Iosevka currently supports.
- [ ] At least two monospace/programming fonts, created by different designers, supported the requested variant. Provide images below.

------

This is a request for a novel variant of zero, with a broken vertical bar through the middle, as shown in the attached image. The primary justification is a matter of accessibility.

I'm unfortunately finding that my vision is getting worse with age. When looking at large quantities of numbers, such as in long lists, tables, or frequently changing counters, especially under poor conditions such as glancing over at a monitor from a distance or at an angle, a common problem arises: A typical slashed or dotted zero is confusable with the digit eight. Both appear as rounded shapes with a weighty middle. An ordinary hollow zero is of course  confusable with a capital O. I needed a zero shape that minimized confusability with any other common letter or digit.

I came up with this solution myself, and have already built Iosevka with it (diffs are attached), though I rather bumbled my way through adding the changes I needed and it's probably not the ideal way to implement it in code. I've been using it for over a month and I find this zero very easy to distinguish at a glance from other digits while still looking presentable up close. I hope this zero variant may be included in the official source to benefit others who must rely on their poor vision. 

![Screenshot1](https://user-images.githubusercontent.com/81892452/233215692-576ea516-2f33-4df5-a814-0b52641977d8.png)

[variants.toml.diff.txt](https://github.com/be5invis/Iosevka/files/11278089/variants.toml.diff.txt)
[0.ptl.diff.txt](https://github.com/be5invis/Iosevka/files/11278095/0.ptl.diff.txt)

be5invis:
This is interesting. Do you know other fonts doing this?

Logo121:
Not vertical, but Fixedsys's 0 is sometimes described as a "broken slash", though not exactly
![](https://upload.wikimedia.org/wikipedia/commons/thumb/e/e6/Fixedsysfontpreview.svg/220px-Fixedsysfontpreview.svg.png)

Might not be a justification to this variant, but I think this alternative also tackles the problem OP has?

Lieutenant-L-T-Smash:
> This is interesting. Do you know other fonts doing this?

[...]

Issue #1509, opened 2023-01-09
Taller 0 slash from v2.x?


nathanielevan:
- [x] The requested variant shape does not go too far away from Iosevka's design.
- [x] The requested variant does not conflict with any characters in Unicode that Iosevka currently supports.
- [ ] At least two monospace/programming fonts, created by different designers, supported the requested variant. Provide images below.

------

Hi, many thanks for your work on Iosevka, been using it for months and loving it! A little question, I came across the older v2.x versions of Iosevka and found out that in that older version, the slash on the 0 is a little taller, like so:

- Older (v2.x): 
![scshot-090123-191755](https://user-images.githubusercontent.com/9361126/211306416-4087afb2-d66c-4538-8c4b-9f80e10e047f.png)
- Current: 
![scshot-090123-191826](https://user-images.githubusercontent.com/9361126/211306447-52b5d4dd-b7d2-40e2-af1f-986e25043b97.png)

And this might seem a bit minor, but I'm wondering if I can get the 0 from the older version with a certain build parameter? Or perhaps, would you consider adding that as a new variant?

Edit: did a little more digging and I found that version [3.0.0-beta.3](https://github.com/be5invis/Iosevka/releases/tag/v3.0.0-beta.3) was the last version to carry the taller 0 slash. The following version, [3.0.0-rc.1](https://github.com/be5invis/Iosevka/releases/tag/v3.0.0-rc.1), seems to no longer ship with that taller 0 slash. It was probably changed with commit [0bac9a1](https://github.com/be5invis/Iosevka/commit/0bac9a13dd5d3cb38e9e2230f73a2c6df3efdc66), but I failed to compile this version i.e. this repository at this commit. Version commit [34307b6](https://github.com/be5invis/Iosevka/tree/34307b6e8554a2281ef36b0cd3cd9c57c23162c9) is the nearest commit after it that compiles.

be5invis:
Do you want more slope for the inner bar, or changing the metrics of the circle?

nathanielevan:
More slope for the inner bar, I think! I'm not a font expert but I can't seem to make out any difference in the metrics of the circle between the two versions -- the slope is the only thing that stands out to me

be5invis:
It will be added in version 24. Resolve as fixed.

Issue #1307, opened 2022-02-15
Interesting variant of zero: broken zero


wimstefan:
- [x] The requested variant shape does not go too far away from Iosevka's design.
- [x] The requested variant does not conflict with any characters in Unicode that Iosevka currently supports.
- [ ] At least two monospace/programming fonts, created by different designers, supported the requested variant. Provide images below.

------

Yesterday I saw a very interesting and visually appealing variant of zero: a broken zero
Pictures say more so I've created an example with Gimp:
![broken-zero](https://user-images.githubusercontent.com/4605934/154027279-29a4ce63-fec0-4538-97df-bb0bf08deccf.png)

I was searching around a bit to find other monospaced/programming fonts using that already but didn't succeed. I find it too pretty though not to suggest it :)

be5invis:
It reminds me _Fälschungserschwerende Schrift_(German license plate font)'s `0`:
![image](https://user-images.githubusercontent.com/240091/154183720-67847d6d-be66-4d6b-987c-55e6d65dfbf6.png)
But this ultimately looks very alien to the designs of `0`. Mark as Needs Justification.

wimstefan:
For me it is a very elegant way to discern the `0` from `capital o` 

There are already broken/open lines in other variants like `capital p`, `capital q`,  `capital r` and of course the `dollar` sign that I'm already using. It's visually very appealing to me - but that's of course a very personal perception.

At the end it is completely up to you whether you decide to implement this variant - either way I'm more than grateful for this wonderful font that you've created :)

akarzim:
If this variant were to exist, I would be more than happy to use it for italics where I already use a large amount of open characters :)

wimstefan:
Hmm I can't think of more justifications to implement this feature but show it in context with other broken glyphs and most of all  `capital o` to make the visual effect a bit clearer:
![broken-numbers](https://user-images.githubusercontent.com/4605934/155848596-67910fee-aaab-4951-9cae-5c4cf250250e.png)

Position, angle & thickness of the broken line can of course be optimized - this is just a raw sketch in Gimp ...

wimstefan:
And I find the solution of *Fälschungserschwerende Schrift* also very enticing ...
![broken-numbers](https://user-images.githubusercontent.com/4605934/155888501-a72b961c-e750-4f80-926d-dfb8de8f70eb.png)

paolo-bolzoni:
[...]

PR #1734, opened 2023-05-19
(Draft) Testing a few backlogged Zero variants (+ a few others)

Logo121:
Starting this because the major number is going to change anyway:
image

This covers the proposed zero variants suggested in:
* #1307:
  * Slash cutout with same slope as the default slashed 0 (https://github.com/be5invis/Iosevka/issues/1307#issue-1138378807)
  * Slash cutout with steeper slope (https://github.com/be5invis/Iosevka/issues/1307#issuecomment-1052195735)
  * Top-right cutout, like the license plate font (https://github.com/be5invis/Iosevka/issues/1307#issuecomment-1053588047)
  * Vertical cutout, like stencil font (https://github.com/be5invis/Iosevka/issues/1307#issuecomment-1553649962)
* #1509
  * Slash with steeper slope (from (0, 0) to (Width, top), I assume)
    * This is aligned with the steeper-slope cutout above
* #1678 
  * Broken vertical bar
  * Broken Slash, like Fixedsys, as I've mentioned and demonstrated (https://github.com/be5invis/Iosevka/issues/1678#issuecomment-1518214429)
    * The slope is aligned with that of Empty Set, because there would be very little remaining if the other slopes are used

Plus their respective Oval/onum forms. The config has also been modified to support the "broken" shapes.

This is just for testing now, and I don't think *all* of these (especially those raised by me) are to be added anyway. For this reason, I also didn't make the reversed slash equivalent, or edit the variant table. These are all encoded in the PUA (U+EF00...U+EF1B) for now.

This is how it looks like zoomed out: image

I'm putting this here to know what should be done next, in particular:
* Which to keep, or even what to add (e.g. reversed variants)
* What optimizations/fixes should be done
* How to order them with the existing zero variants


wimstefan:
Thank you so much for all the hard work you put into it!
It would be simply fantastic to see my cautious request being realized after such a long time 😃

font-src/glyphs/number/0.ptl
```diff
@@ -31,16 +34,65 @@ glyph-block Digits-Zero : begin
...
+	define [CutoutSlash top] : dispiro
+		flat 0 (top * (1 - 0.77)) [widths.center (top * 0.125)]
+		curl Width (top * 0.77)
+
+	define [CutoutTallSlash top] : dispiro
+		flat 0 0 [widths.center (top * 0.125)]
+		curl Width top
+
+	define [CutoutVerticalBar top] : dispiro
+		flat Middle (top - O) [widths.center (Width * 0.125)]
```
be5invis, line 60: Change the stroke width after (`widths.center`) to things like
`[Math.min (0.4 * CircleInnerWidth) [AdviceStroke 4]]`

font-src/glyphs/number/0.ptl
```diff
@@ -31,16 +34,65 @@ glyph-block Digits-Zero : begin
...
+			include : difference zero overlay
+			include zero
+
+
+	define [CutoutSlash top] : dispiro
+		flat 0 (top * (1 - 0.77)) [widths.center (top * 0.125)]
+		curl Width (top * 0.77)
+
+	define [CutoutTallSlash top] : dispiro
+		flat 0 0 [widths.center (top * 0.125)]
```
be5invis, line 56: Change the stroke width after (`widths.center`) to things like
`[Math.min (0.4 * CircleInnerWidth) (top * 0.15) [AdviceStroke 3]]`

Adjust the parameters to make the SW being proper under all the weights.

font-src/glyphs/number/0.ptl
```diff
@@ -31,16 +34,65 @@ glyph-block Digits-Zero : begin
...
+			include overlay
+
+	define [Cutout zero overlay fnMask height] : glyph-proc
+		if overlay
+			include : difference zero overlay
+			include zero
+
+
+	define [CutoutSlash top] : dispiro
+		flat 0 (top * (1 - 0.77)) [widths.center (top * 0.125)]
```
be5invis, line 52: Change the stroke width after (`widths.center`) to things like
`[Math.min (0.4 * CircleInnerWidth) (top * 0.15) [AdviceStroke 3]]`

Adjust the parameters to make the SW being proper under all the weights.

font-src/glyphs/number/0.ptl
be5invis: Adjust the parameters to make the SW being proper under all the weights.

be5invis (changes requested):

be5invis:
And well, you can keep all the variants (modify `variants.toml` to add them under `[prime.zero.variants]`). Since V23 will be a major version change you can freely reorder the variants' rank.

Lieutenant-L-T-Smash:
+1 to the steeper slash. If anything makes it through to final, it must be that one. Looking at this chart and being able to compare, it's surprising even to me how much difference that makes in making it distinguishable from the 8 under astigmatic vision.

The vertical cutout just looks like a pair of parentheses to me.

Logo121:
Ok, I tried making the gap vary a bit more, though if I just use AdviceStroke without a lower bound, the gap will be almost non-existent in Thin (and Extralight to some extent): image

I tried to keep the effect that the gaps are still discernible when zoomed out (or at around 8pt), at least for me:
imageimage

And this is what it looks like up close (at around 24pt):
image image

Not sure if it's better now.

PR #1753, opened 2023-05-27
Disunify partial differential with lowercase eth

jmcwilliams403:
The unification of Partial Differential and Lowercase Eth was honestly not the best idea. Partial differential is based on a script lowercase delta, and Cyrillic De's italic form derives from the same concept, as Cyrillic is based on the Greek script. This is honestly the more sensible approach, but unfortunately it means that another version bump is inevitable.

### DIFF
--- a/params/variants.toml
+++ b/params/variants.toml
@@ -5926,10 +5926,27 @@ selector.cent = "barInterrupted"
 
 
 
+[prime.partial-derivative]
+sampler = "∂"
+samplerExplain = "Partial derivative symbol"
+tag = "VXSF"
+
+[prime.partial-derivative.variants.straight-bar]
+rank = 1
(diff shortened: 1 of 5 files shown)


params/variants.toml
```diff
@@ -5926,22 +5926,20 @@ selector.cent = "barInterrupted"
 
 
 
-[prime.partial-derivative]
-sampler = "∂ ð"
-samplerExplain = "Partial derivative symbol and Lowercase Eth"
+[prime.lower-eth]
```
be5invis, line 5929: We should be able to create _two_ prime variants: one for eth and insular-d, one for partial derivative sign. The Eth variant will be assigned to tag `VXAA`. This could avoid creating another major version.

be5invis (changes requested):

be5invis (approved):

PR #1780, opened 2023-06-07
Improve mark placement of small H with diaeresis

jmcwilliams403:
```
ḧĥȟ
ḣḃḋ
```
image


be5invis:
I do not think this visual is good, as dieresis is much wider so it looks very unbalanced.

jmcwilliams403:
Should we just close this PR then? It's entirely cosmetic but if the aesthetics are bad then it might not be worth it.

PR #1782, opened 2023-06-07
Some curly tail characters

Logo121:
image
* Adds some of the curly tail (or "crossed-tail") characters. The superscript `𐞺` completes Latin Ext F.
* Modified the CurlyTail code so that the `right` position is a bit more logical. This affects all other existing curly tail characters, but not too much to cause any distortion (in fact, a few of them extrude less outside the boundary after the change)
* `ꬶꬼ` kinda based on j with curl (`ʝ`), but the curl squashed to prevent overlapping with the body.
* `𝼞` based on alveopalatal c (`ɕ`). Same adjustment as above.
* Originally I tried to align `ꬻꬺ` with `ȵ`, but realized that the curl for cross-tail n would look to squashed in thin weight. They are aligned by XH for now.
  * In Noto Sans Mono, the curl point slightly downward (achievable by removing `O` in the `y2` parameter). But since most other fonts (including the Unicode sample) point horizontally, I follow that version.
  * I'm really unsure about `ꬺ`: The current setting tries to balance between 
    1. Preventing overlaps (other than the cross)
    2. Preventing the start of the curl from bending weirdly due to the `mid` variable shifting too far right (still slightly visible in `ꬻ`) image
    3. Preventing NaN angles due to the curl gap being too small
  
    But it looks a bit thin as a result. Maybe shifting the middle bar to the left is a solution, but since only unifont does that, I'm not sure if I should.
  * The inner serif of `ꬻ` will be omitted if the curl is too close (or overlaps) it. Same should happen for `ꬺ`, though I'm not sure if my implementation is correct.
* For `ꬶꬼ𝼞`, The `right` position is based on where the stroke will normally end without the curl (assuming default variant).
* Works with variants: image

### DIFF
--- a/font-src/glyphs/letter/latin/s.ptl
+++ b/font-src/glyphs/letter/latin/s.ptl
@@ -15,11 +15,12 @@ glyph-block Letter-Latin-S : begin
 	glyph-block-import Letter-Shared-Shapes : ArcStartSerifWidth ArcStartSerifDepth
 	glyph-block-import Letter-Shared-Shapes : SerifedArcStart SerifedArcEnd
 	glyph-block-import Letter-Shared-Shapes : ArcStartSerif ArcEndSerif
-	glyph-block-import Letter-Shared-Shapes : PalatalHook RetroflexHook CyrDescender
+	glyph-block-import Letter-Shared-Shapes : PalatalHook RetroflexHook CyrDescender CurlyTail
 
 	define SLAB-NONE      0
 	define SLAB-CLASSICAL 1
 	define SLAB-INWARD    2
+	define CURLY-TAIL     10
 
 	define SOBot  OX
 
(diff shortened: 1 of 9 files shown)


font-src/glyphs/letter/latin/s.ptl
```diff
@@ -41,6 +42,15 @@ glyph-block Letter-Latin-S : begin
 			alsoThru.g2 0.5 0.5 [widths.center ess]
 			g4.down.mid (deltaX + RightSB - Width - SOBot) (bot + archDepth) [widths.rhs gap]
 
+	define [SCurlyTail sw] : begin
+		local fine : AdviceStroke2 3 4 XH
+		local rinner : ((XH * 0.5) - fine * 2 - sw * 0.5) / 3
+		local m1 : RightSB - SOBot - sw * HVContrast
```
be5invis, line 48: Could try decrease this so the right terminal of the tail won't hit the body of `s`.

Logo121:
Not sure what you wanted me to decrease (or where the body of `s` ends), but here's one with `rinner` changed to `((XH * 0.5) - fine * 2 - sw * 0.5) / 4`:
image
And here's one with `x2` shifted leftwards:
image

The first one is closer(?) to the Unicode sample and Gentium (and Nishiki-Teki), while the second one is closer to Noto Sans and Serif.

Which one is better?

be5invis:
@Logo121  First. I will merge now.

Issue #1024, opened 2021-05-19
U+A7F8 MODIFIER LETTER CAPITAL H WITH STROKE should be small capital


dscorbett:
According to [L2/20-038](https://www.unicode.org/L2/L2020/20038-voqs-req.pdf), U+A7F8 MODIFIER LETTER CAPITAL H WITH STROKE is really a superscript _small_ capital H with stroke. Cf. U+02B6, which is the correct size.
![Vꟸʶ](https://user-images.githubusercontent.com/1124347/118897392-251c7200-b8d8-11eb-91bf-cd2575f773eb.png)

---

* Your font version: 7.0.0-preview.2
* Your font variant: Iosevka Slab Extended
* Your operating system (name and version): macOS 10.14.6
* Your application using Iosevka: hb-view

PR #1798, opened 2023-06-17
I Longa

jmcwilliams403:
also cleanup duplicate Latin upsilon with stroke binding which was my fault in the first place.
`ɪIꟷꟾ̬̂ʊᵿ`
image

### DIFF
--- a/changes/24.1.4.md
+++ b/changes/24.1.4.md
@@ -1,3 +1,5 @@
+* Add Characters:
+  - LATIN EPIGRAPHIC LETTER I LONGA (`U+A7FE`).
 * Fix shape of `U+1D95` (#1790).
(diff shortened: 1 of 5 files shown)


dscorbett:
The current shape of U+A7F8 MODIFIER LETTER CAPITAL H WITH STROKE is correct; see #1024.

I longa should be shifted up instead of having a descender.

jmcwilliams403:
> The current shape of U+A7F8 MODIFIER LETTER CAPITAL H WITH STROKE is correct.

Undone just now.

> I longa should be shifted up instead of having a descender.

There really isn't much room for a capital letter with ascender. Plus, this is how Microsoft fonts draw it. I'll leave it as is for now, and allow for future improvement in case anyone else wants to try.

PR #1813, opened 2023-06-24
circled symbols with backslash

Logo121:
Full width and Half width:
image
Basically using the same slash as the backwards slash zero.

Works with variants:
image

Also, since these, along with circled zero and CC are supposed to be in the "same set" of symbols as the copyright/copyleft ones, I suppose they would all be `FullWidth2`.

Apparently the "no derivatives" symbol got unified into Circled Equal `⊜` so a variant reflecting that would be cool, but eh. ~~No font I know does that.~~ A few fonts do so, but they are either [really scoped to just support CC symbols](https://github.com/publicdomain-nocopyright/Hosted-Public-Domain-font-with-symbols), or [decides to support ⊜ as not a math symbol](https://www.fontspace.com/retrcade-font-f88842)


Logo121:
Also fixing #1814 just for the sake of it
image

jmcwilliams403:
A lot of other symbols are used as license symbols like (Wz), (CD), etc. but the problem with compressing those to be narrow by default is that you pack so much geometry into a single character cell that it reduces readability. I think you could only really get away with it for characters that _don't specifically_ have a unicode name of e.g. "circled [fill in the blank]"

Logo121:
The problem is when the CC symbols are used they are really used alongside each other, so at least they should have a consistent size.

[Use case 1](https://games.porg.es/articles/cards/japan/hanafuda/intro/)
image
[Use case 2 with Circled Equal](https://ja.sntmse.com/)
image
(on my end it uses Iosevka for CC and Noto for the rest I think, I'm not sure)

I don't know much cases (other than Unifont but well, it's Unifont) where a font has (C) in Halfwidth, but the CC symbols in fullwidth. While for (Wz) and (CD), they are not always rendered the same.

* Fairfax has all half (including Wz CD, but a cropped circle): image
* Nishikiteki is not monospaced but has them with the same size and enclosing circle: image
As for (Wz) and (CD), they are rendered like the other enclosed alphanumerics. (Wz) in particular is oval like in Unicode. image
* PragmataPro has all full (from repo): image vs image, though no support for (Wz) and (CD)

jmcwilliams403:
Fair enough. Perhaps your approach here is better then, as it's both taking into account which characters were added with the original intention of being a license symbol and also which characters' recommended glyphs are circular or oval.

PR #1820, opened 2023-06-27
Style set corrections, kappa variant assignments

jmcwilliams403:
Firstly, some corrections to stylistic sets' variation assignments are made based on the source fonts.

Secondly, in a large variety of fonts, latin Kra and greek Kappa often follow a slightly different serifed form from small capital K.
Example from [Wikipedia](https://en.wikipedia.org/wiki/Kra_(letter)):
image

Additionally, Latin Capital Chi should follow Capital X similarly to how Latin Capital Beta follows Capital B.

Below is a test using the default slab configuration.

From left to right:
Latin Capital K, Latin Kra, Latin Small K, Greek Kappa, Latin Capital X, Latin Capital Chi, Latin Small Chi, and Greek Chi.
`KĸkκXꞳꭓχ`

Regular:
image
Italic:
image

### DIFF
--- a/params/variants.toml
+++ b/params/variants.toml
@@ -525,7 +525,6 @@ descriptionAffix = "straight shape"
 selectorAffix.K = "straight"
 selectorAffix."K/sansSerif" = "straight"
 selectorAffix."grek/KaiSymbol" = "straight"
-selectorAffix."grek/kappa" = "straight"
 selectorAffix.KDescender = "straight"
 
 [prime.capital-k.variants-buildup.stages.body.curly]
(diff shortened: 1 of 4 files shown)


params/variants.toml
```diff
@@ -2361,6 +2362,8 @@ selectorAffix.k = "cursiveTailed"
 selectorAffix."k/sansSerif" = "cursiveTailed"
 selectorAffix.kHookTop = "cursiveTailed"
 selectorAffix.kDescender = "cursive"
+selectorAffix."latn/kappa" = "curly"
+selectorAffix."grek/kappa" = "curly"
```
be5invis, line 2366: Same as above.

be5invis (changes requested):

be5invis:
May also think about "de-cursive-ing" of `x` to be straight instead of curly.

PR #1824, opened 2023-07-02
Eta and Insular R fixes

jmcwilliams403:
Most fonts have Greek Small Eta follow lowercase N, most noticeably on Ubuntu mono which inherits its earless cornered shape.

Additionally, using a full retroflex hook for Insular R puts both legs too close to each other. The left leg should descend below the baseline, but the right leg should not.

`Ꞃꞃnηῃ`

Sans:
image

Slab:
image

`ss12`:
image

Original Ubuntu Mono font:
image

### DIFF
--- a/font-src/glyphs/letter/latin/lower-n.ptl
+++ b/font-src/glyphs/letter/latin/lower-n.ptl
@@ -142,6 +142,13 @@ glyph-block Letter-Latin-Lower-N : begin
 			if sLB : include : sLB 0
 			if sRB : include : sRB Descender
 
+		create-glyph "grek/eta.\(suffix)" : glyph-proc
+			include : MarkSet.p
+			set-base-anchor 'lf' (SB + HalfStroke) 0
+			include : Body XH SB RightSB Descender Stroke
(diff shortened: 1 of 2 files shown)


font-src/glyphs/letter/latin/lower-n.ptl
```diff
@@ -174,16 +181,16 @@ glyph-block Letter-Latin-Lower-N : begin
 
 		create-glyph "RInsular.\(suffix)" : glyph-proc
 			include : MarkSet.capDesc
-			include : Body CAP SB RightSB (0 + TailY - HalfStroke) Stroke
-			include : RetroflexHook.rExt RightSB (0 + TailY - HalfStroke)
+			include : Body CAP SB RightSB (XH - SmallArchDepthB + O) Stroke
```
be5invis, line 184: We *should* use the full `RetroflexHook` for tails that carry semantics.
Just change the height used here to `(0 - descender - HalfStroke)`.

be5invis (changes requested):

jmcwilliams403:
Current:
image

PR #1844, opened 2023-07-14
Some Christian Crosses

### LINKED ISSUE #1841: Add U+271D ✝ (Latin Cross)
  - [ ] Latin
  - [ ] Cyrillic
  - [ ] Greek
  - [ ] Punctuation
  - [x] Symbol
- [ ] Some other monospace/programming fonts supported this character. Provide images below.

------

This symbol is used by [Lean4](https://github.com/leanprover/lean4) to represent variables whose name is irrelevant (no clue why, but the abbreviation for it `\textdied`). Since Iosevka doesn't currently support this character, my editor falls back to using it as an emoji, which looks very out of place, so adding this would be greatly appreciated.

Logo121:
Resolves #1841, with a few more related glyphs. Using the same approach to other Geometric mosaic characters (in particular Greek Cross), implemented like a Greek Cross but the bottom bar is longer than the other 3.

`a`, Dagger and Heavy Greek Cross for reference: 
image

The issue discussed about differentiation with Dagger, but it turns out Iosevka's Dagger is so tall anyway that it is probably different enough already.

It is Fullwidth by default though, following Greek Cross. Here's the halfwidth version:
image

### DIFF
--- a/changes/25.0.2.md
+++ b/changes/25.0.2.md
@@ -1,3 +1,9 @@
+* Add Characters:
+  - ORTHODOX CROSS (`U+2626`).
+  - CROSS OF LORRAINE (`U+2628`).
+  - LATIN CROSS (`U+271D`) (#1841).
+  - HEAVY LATIN CROSS (`U+1F547`).
+  - CELTIC CROSS (`U+1F548`).
 * Refine shape of `U+A666` and `U+A667` (#1838).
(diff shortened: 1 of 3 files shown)


be5invis:
The Lorraine and Celtic cross needs more optimization in Wide.
image

Logo121:
hmm

Is it fine if I just make the cross larger (using the Size.Large scale for height)?  
image
image

PR #1877, opened 2023-07-28
Circular Arrows, CC-BY and SA

Logo121:
The remaining 2 CC Symbols (with copyleft for reference):
imageimage
* Tried my best to recreate the official CC-BY symbol human figure, and making it depend a bit on weight as well (i guess?)
* For the CC-SA arrow, [wiktionary](https://en.wiktionary.org/wiki/%F0%9F%84%8E#Translingual) claims that it is a variant of the copyleft symbol but with an arrowhead. So I did that instead of using the original CCW arrow style.
  * I suppose we *can* make the arrow fullwidth in WWID (closer to the original) but that would be kinda weird without syncing copyright and copyleft as well. I'll just leave it as is.

I suppose that completes the set, if ignoring `⊜`.

----

The circular arrows (which I did in preparation of the ShareAlike arrow. Default is Wide):
imageimage
* Redid Semicircle Arrows (`↶↷`) and Open Circle Arrows (`↺↻`) to kinda match how Iosevka handles "Halfwidth Circular Symbols" like faces or encircled letters (or to mimic PragmataPro's style). Also so that it generalizes better to the other circular arrows.
   * Unfortunately this forced me to bump the stroke weight to `[AdviceStroke 5]` due to the halfwidth open circle arrow heads colliding with the stem.
* Triangle-Headed Open Circle Arrows (`⭮⭯`) are (probably) just the open circle arrows but with triangle head. Most font shows these (and the original Circle Arrows) with an opening above, but I'll follow the existing style for now.
* Closed Circle Arrows (`⥀⥁`) are open circle arrows with the ring closed.
* Gapped Circle Arrows (`⟲⟳`) have two common renditions: A 1/8-open circle, or a circle with a tiny gap (kinda like `∲`). I chose the latter because it's probably easier to do (considering the half-width version).

### DIFF
--- a/changes/26.0.0.md
+++ b/changes/26.0.0.md
@@ -3,10 +3,18 @@
 * \[**Breaking**\] Disunified the variant selector for Greek Delta and Greek Lambda, and added selectable serif variants for Lambda (#1866).
 * \[**Breaking**\] Add a separate variant selector for lowercase Greek Chi.
 * Add Characters:
+  - ANTICLOCKWISE GAPPED CIRCLE ARROW (`U+27F2`).
+  - CLOCKWISE GAPPED CIRCLE ARROW (`U+27F3`).
+  - ANTICLOCKWISE CLOSED CIRCLE ARROW (`U+2940`).
+  - CLOCKWISE CLOSED CIRCLE ARROW (`U+2941`).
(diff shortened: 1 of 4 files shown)


nsemrau:
@Logo121 My day-job entails publishing works under various CC licenses. Because of that, I just wanted to thank you for your work on Iosevka in general, and thank you for the CC buttons in Iosevka especially.

Issue #1067, opened 2021-05-26
'cv29' does not apply to all variants of ⟨d⟩


dscorbett:
U+0256 LATIN SMALL LETTER D WITH TAIL and U+1D91 LATIN SMALL LETTER D WITH HOOK AND TAIL are not affected by 'cv29'. For example, here is 'cv29' = 2.
![dɖᶑ](https://user-images.githubusercontent.com/1124347/119585178-2a763280-bd98-11eb-91c4-89bb380e758d.png)

---

* Your font version: 7.0.0-preview.5
* Your font variant: Iosevka Slab Extended
* Your operating system (name and version): macOS 10.14.6
* Your application using Iosevka: hb-view

be5invis:
Variants for `U+1D91` don't make sense, since all modifiable parts (top serif, bottom-right-corner) are occupied by hooks or tails.

dscorbett:
There is also the transition (corner or rounded) between the bottom of the bowl and the vertical stroke: that still makes sense despite hooks and tails.

PR #1890, opened 2023-08-03
Reduce the amount of variants used by d/p/q with hook/tail

jmcwilliams403:
The versions of `d` with a tail are only marginally different when toothless, and `p`/`q` with hooks are only marginally different when earless.
All it previously did was add a tooth or ear back onto a toothless or earless variant, which lead to some heavy intersections between the toothless/earless corners and the hook extensions.

What this effectively does is forces the variants used by these characters to be toothed or eared, effectively putting all these characters in harmony with each other.

Something to note is that now `ᶑ` is exactly the same across all variants.

```
dᶁuᶙ
ɗɖᶑ𐞌𐞋𐞍
pƥqʠ
```
Default Slab:
image
Default Slab Italic:
image
`ss12`:
image
`ss12` Italic:
image


be5invis:
I used to do this for `d` but @dscorbett raised #1067 so I explicitly allowed these additional variants. Per David:

> There is also the transition (corner or rounded) between the bottom of the bowl and the vertical stroke: that still makes sense despite hooks and tails.

be5invis:
One thing we _could_ do is perhaps limiting this to only earless-rounded / toothless-rounded. For the cornered variant just keep it as-is, as it forms a T-junction.

jmcwilliams403:
In this PR's current state this is not salvageable. I might return to this idea some other time.

PR #1891, opened 2023-08-03
Volapuk Letters, and a few other fixes

### LINKED ISSUE #1865: Latin Ext. D: Volapük AE, OE, and UE (`U+A79A`..`U+A79F`)
  - [x] Latin
  - [ ] Cyrillic
  - [ ] Greek
  - [ ] Punctuation
  - [ ] Symbol

------

These letters are used as typographic substitutes for Ää/Öö/Üü in the Volapük language and also see some use in phonetic transcription of Germanic dialects.

`Ꞛꞛ Ꞝꞝ Ꞟꞟ`

Noto Sans Mono:
image

Fairfax HD:
image

Everson Mono:
image

Logo121:
* The "gap" of the share-alike arrow was a bit off-center. (`arrow.ptl`)
* Added Single-storey forms for Iotified-a, seeing how Brill and Noto (proportional) had them anyway. (`iotified-a.ptl`)
  * This meant that the SingleStorey `a` functions also had to be extended to support DivFrames and custom Stroke width (`lower-a.ptl`).
Hopefully that didn't break anything (`aąɑⱭꭤꙗ` in `cv26 = 08, 10, 12`, then italic):
image
* Some other files have minor name/formatting fixes as I briefly went through them (`c.ptl`, `transformed.ptl`, `orthography.ptl`)

----

Volapuk Letters, closes #1865 (which should've been a separate PR, but oops)
image
Implemented by extending Epsilon code (`lower-epsilon.ptl`) by quite a lot. I kinda made a mess...

* Supports variants (see below)
* Basically divided the glyphs into possible upper-lower shape component patterns (including variants), including:
  * Parts that open vertically: image
  * Parts that end up in a stem, like the OBar shapes: image
These differ a bit from the existing `SLAB-CLASSICAL` shape, which I had to add a new function in `shared.ptl` for (in OBarLeft/Right, because idk where to put them)
  * Parts that connects toothlessly(?) to a stem: image
These are identical to `SLAB-INWARD` as far as I'm aware.
  * Parts that connects rounded to a stem: image
  * Parts that connects like into OShape, which only differs from the previous one by an `OX` on the terminal side.
* While I'm at it I also adapted Komi Dzje and Zje to this format instead of having a "Half Lower Shape", as well as the closed Epsilons in `upper-b.ptl`.
  * I don't know if I should even do the same for Beta symbol or cursive Cyrillic ve. I just left it untouched for now.
* Changes to other modules for these glyphs:
  * A new variant type `a/single` because `scripta` does not allow ear variants. (`variants.toml`)
  * Grouping, renaming and adding `sw` support for U Serifs (`u.ptl`), for better access in the epsilon module.
* This might not be super important, but since the left side of these glyphs are not actually epsilons but "dents" to the right, they are differentiated by the depth of the middle part: image

AE with `cv26 = 08, 10, 12` then italic, compared to Script A:
image
UE with `cv45 = 03, 10, 12` then italic, compared to Lower U:
image
OE compared to O, for completeness sake:
image

----
As a final note, here's some minor bugs I'm not able to fix:
* These small bulges in the "vertical" parts, in heavier weights: image
* The arc extruding the serif for inward-serif Cyrillic ze, in Heavy (exists before this PR): image

### DIFF
--- a/changes/26.0.0.md
+++ b/changes/26.0.0.md
@@ -18,6 +18,7 @@
   - LATIN SMALL LETTER K WITH DIAGONAL STROKE (`U+A743`).
   - LATIN CAPITAL LETTER K WITH STROKE AND DIAGONAL STROKE (`U+A744`).
   - LATIN SMALL LETTER K WITH STROKE AND DIAGONAL STROKE (`U+A745`).
+  - LATIN CAPITAL LETTER VOLAPUK AE (`U+A79A`) ... LATIN SMALL LETTER VOLAPUK UE (`U+A79F`) (#1865).
   - CIRCLED ANTICLOCKWISE ARROW (`U+1F10E`).
   - CIRCLED HUMAN FIGURE (`U+1F16F`).
 * Drop `<=` and `>=` as inequality for Verilog (#1864).
@@ -29,3 +30,4 @@
 * Fix variant selection for `cv27`, `cv33`, `cv36`, and `cv49` for `ss17` under italics.
 * Make Greek Kappa respond to top-left serifed variants of `k`.
 * Fix slabs for `U+019C`, `U+0257`, `U+026F`, and `U+0270`.
+* Add single-storey variants support for `U+A657`.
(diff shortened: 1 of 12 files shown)


be5invis:
Good Job.

PR #1903, opened 2023-08-09
Make Etoile more closely resemble serif form of Aile

jmcwilliams403:
Also fix serifs of `w` under italics to match default slab.

Etoile:
image

Compared to Aile:
image

Etoile Italic:
image

Aile's faux italic form for completion sake:
image


be5invis:
Etoile's style choice is NOT designed to become the serifed version of Aile. They do NOT need to match.
Perhaps we only need the fixes for `w`

PR #1919, opened 2023-08-13
Add Volapük sample, update Dutch sample

jmcwilliams403:
Using Volapük's specific Ä/Ö/Ü surrogate characters.

Also, the Dutch `ĳ` ligature is _supposed_ to take up one character cell in normal typesetting. It's often considered its own letter, had its own typehead on typewriters, and has its own tile in games of Scrabble.
I understand, however, why the Croatian `ǅ` etc. digraph characters were considered unnecessary; They're for round-trip compatibility with Cyrillic, not strictly orthographic, and thus make no sense to pack into one cell.


Logo121:
Aren't the volapuk letters barely used though? Literally the only usage of them I could find is the name Volapuk itself (or like 1 or 2 people advocating the use of it again).

jmcwilliams403:
I just think that using the umlauted letters makes it redundant to the German sample. Its optional vowel surrogates are a better showcase of glyphs unique to the language.

PR #1929, opened 2023-08-16
Modifier Glottal Stop adjustment and a couple slab variant selection fixes

jmcwilliams403:
Raised glottal stop:
`xˀ𐞳ˁ𐞴ˤ`
image

Also, this stylistic set didn't have a slab override until recently, so I fixed the italic slab variants of any characters previously unaccounted for:

`fkvwxyzſß`

`ss18` slab:
image
`ss18` slab italic:
image


dscorbett:
The modifier letters are already the right size. It might be okay to change U+02C0 and U+02C1, but not U+02E4. In version 26.1.0 they look like:
ˀ𐞳ˁ𐞴ˤ

U+02C0 and U+02C1 form a pair. According to [the code chart](https://www.unicode.org/charts/PDF/U02B0.pdf), U+02C0 is a “typographical alternate for 02BC ʼ or 02BE ʾ” and U+02C1 is a “typographical alternate for 02BF ʿ” and they don’t look exactly like superscript U+0294 and U+0295. That is, they are more like modifier _symbols_ using the glottal stop shape than modifier _letters_ in the usual sense, even though they are categorized as modifier letters.

U+107B3, U+107B4, and U+02E4 are normal modifier letters with \<super\> decompositions. “SMALL” in the name of U+02E4 just distinguishes it from U+02C1, which already took the expected name “MODIFIER LETTER REVERSED GLOTTAL STOP”. It means it is based on a lowercase letter, not that it is smaller than U+02C1.

In [L2/20-046 § 1b](https://www.unicode.org/L2/L2020/20046-script-adhoc-rept.pdf#page=3), the Script Ad Hoc group opined that U+02C0 could be used as a normal superscript glottal stop. In that case, it would be fine to change it as this PR does. It might also be fine to make the analogous change to U+02C1, as this PR does, but I am not sure.

jmcwilliams403:
Also, there exist a lot of other fonts who don't agree with each other on whether Modifier Reversed Glottal Stop or Modifier Small Reversed Glottal stop should be larger or smaller than the other. I can try to find examples.

jmcwilliams403:
I stand corrected. I found more examples matching how it is currently. I'll draft this, undo that part, and reopen

jmcwilliams403:
Kinda funny how my original change to Glottal Stop in #1706 was actually correct, but then the new change I tried here was a step backward. Something something "a broken clock is right twice a day".

dscorbett:
Should U+02C0 and U+02C1 have bottom serifs? In the code chart and in some serif fonts, they don’t, treating them more like symbols. Some fonts do give them serifs though.

jmcwilliams403:
Maybe I should make them raised and serifless. Drafting again.

Logo121:
I'd lean towards changing superscript glottal stop if the Unicode committee is not interested in adding a separate character for it.

Out of the 3 fonts [IPA releases their charts in](https://www.internationalphoneticassociation.org/content/full-ipa-chart), all of them have the two characters as superscript letters (`aˀˁˤ`):
image
(Doulos has U+02E4 as the small letter but ignore that for now)
IPAKiel ([paid font](https://www.linguistsoftware.com/liu.htm)): image

Since Unicode (or at least wiki, idk) designates the two characters as the ones used in Superscript IPA, I do wish they would actually look consistent with the base counterparts. [Though they wouldn't be the only inconsistent ones](https://www.unicode.org/L2/L2020/20117-modifier-letter-suppt.pdf):
image
image
(unless we have get an "IPA form" akin to the "APL form"? lol)

jmcwilliams403:
image

jmcwilliams403:
this PR is becoming a complete mess. I'm scrapping this for now. I ended up rushing out these commits and I'm not happy with them anyway.

PR #1988, opened 2023-09-08
Breakup `R` motion-serifed variant into multiple variants, also add them for Cyrillic Ya

jmcwilliams403:
`R₹ꝚꝛKkκКкЯя`
`ss15`:
image
Compared to IBM Plex Mono:
image
`ss17`:
image
Compared to Recursive Sans Mono Linear:
image
image

### DIFF
--- a/changes/26.3.3.md
+++ b/changes/26.3.3.md
@@ -1,7 +1,16 @@
 * \[**Breaking**\] Change of variant names:
+  - `upper-r`.`straight-motion-serifed` → `upper-r`.`straight-top-left-serifed`
+  - `upper-r`.`curly-motion-serifed` → `upper-r`.`curly-top-left-serifed`
+  - `upper-r`.`standing-motion-serifed` → `upper-r`.`standing-top-left-serifed`
+  - `upper-r`.`straight-open-motion-serifed` → `upper-r`.`straight-open-top-left-serifed`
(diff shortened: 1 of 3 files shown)


changes/26.3.3.md
```diff
@@ -1,7 +1,16 @@
 * \[**Breaking**\] Change of variant names:
```
be5invis, line 1: If it is breaking, then make the version number 27.0.0, and remove all lines in `variants.toml` containing `nonBreakingVariantAdditionPriority` or `nonBreakingTagForNewVariantSelector`.

be5invis: In addition, _measure_ the impact of glyph quantity of your change.

be5invis (changes requested):

jmcwilliams403:
I can't currently change the version name of the \.md file without breaking parity with the dev branch as of 366f377

jmcwilliams403:
Also I chose the breaking option because I already made a breaking change with #1986 and I figured it would be more effort than it's worth to turn everything so far into a non-breaking change at this point.

jmcwilliams403:
Nevermind I just realized that 26.3.3 is already out. I'll start this commit over later without breaking anything.

be5invis:
@jmcwilliams403 I'd like to merge this now as I decided to create version 27.0.
No need for a new PR -- I will fix everything later.

Issue #1882, opened 2023-07-31
Detected broken geometry when processing mCrossedTail.serifless in custom build


jasyip:
Describe your problem here. Provide pictures and files if necessary.

When building a custom variant, I get this unexpected error:
```
 [ACTION] Create TTF 'dist/lj-iosevka-aile/ttf-unhinted/lj-iosevka-aile-semi-condensedheavyoblique.ttf'
Detected broken geometry when processing mCrossedTail.serifless
Error:  Target Error Error: Error: NaN angle detected.
    at NormalRectifier.arcTo (file:///home/runner/work/ljs-iosevka/ljs-iosevka/font-src/support/geometry/spiro-expand.mjs:129:11)
    at collectSpiroArcs (/home/runner/work/ljs-iosevka/ljs-iosevka/node_modules/.pnpm/spiro@3.0.0/node_modules/spiro/lib/index.js:448:12)
    at Module.spiroToArcsOnContext (/home/runner/work/ljs-iosevka/ljs-iosevka/node_modules/.pnpm/spiro@3.0.0/node_modules/spiro/lib/index.js:480:16)
    at SpiroExpander.initializeNormals (file:///home/runner/work/ljs-iosevka/ljs-iosevka/font-src/support/geometry/spiro-expand.mjs:21:11)
    at DiSpiroGeometry.expand (file:///home/runner/work/ljs-iosevka/ljs-iosevka/font-src/support/geometry/index.mjs:153:12)
    at DiSpiroGeometry.asContours (file:///home/runner/work/ljs-iosevka/ljs-iosevka/font-src/support/geometry/index.mjs:126:29)
    at CombineGeometry.asContours (file:///home/runner/work/ljs-iosevka/ljs-iosevka/font-src/support/geometry/index.mjs:337:25)
    at TransformedGeometry.asContours (file:///home/runner/work/ljs-iosevka/ljs-iosevka/font-src/support/geometry/index.mjs:268:31)
    at SimplifyGeometry.asContours (file:///home/runner/work/ljs-iosevka/ljs-iosevka/font-src/gen/finalize/glyphs.mjs:104:30)
    at TransformedGeometry.asContours (file:///home/runner/work/ljs-iosevka/ljs-iosevka/font-src/support/geometry/index.mjs:268:31)
[...]

Issue #1287, opened 2022-02-03
More open “5” and “c”


twhb:
Having the bottom left of the “5” curl less upward makes it more distinct from a “6”, especially at small sizes, while still being recognizable. Similarly, having “c” a bit less closed makes it more distinct from “o” while still being recognizable. A more open variant is already available for “g” (and appreciated), and a somewhat similarly edited variant for “t”.

<img width="330" alt="Comparison of 5 and c between Iosevka, DejaVu Sans Mono and Fira Code" src="https://user-images.githubusercontent.com/3160667/152257495-0b7712e4-1b6e-4d92-9bc1-2c9778fcde1c.png">

***

- [x] The requested variant shape does not go too far away from Iosevka's design.
- [x] The requested variant does not conflict with any characters in Unicode that Iosevka currently supports.
- [x] At least two monospace/programming fonts, created by different designers, supported the requested variant. Provide images below.

be5invis:
Mimicking humanist sans-serif is considered out of scope.

twhb:
I’m not 100% clear on what it means for a font to be humanist, but here are a few that seem to me not to be, and also aren’t on Wikipedia’s list of humanist fonts.

<img width="292" alt="Comparison of 5 and c in Iosevka, Courier New, Inconsolata, and Source Code Pro" src="https://user-images.githubusercontent.com/3160667/152296736-5f4af60d-f431-4674-bbdc-334f39967a97.png">

AndydeCleyre:
![Iosevka Term Custom](https://user-images.githubusercontent.com/1787385/157508626-21d1bee4-cb7b-4977-b700-03fda4d01331.png)

susinodan:
Built Iosevka with [.widths.normal] "shape = 610", and [.metric-override] "cap", "ascender", "xHeight" and "parenSize" scaled down by 0.90 (10% shorter glyph size) to make glyph size close to Source Code Pro. Font size 12 for example:
![изображение](https://user-images.githubusercontent.com/906556/169539853-cc766766-fe66-4078-9dc7-18749577c158.png)

And, yeah, in this case I would agree with @twhb - more open "c" would be good, but not as open, as @twhb pictured. Just a little bit. MsPaint "simulation":
![изображение](https://user-images.githubusercontent.com/906556/169539909-880fb78e-1907-4320-902a-26aabac98861.png)

AndydeCleyre:
I'm still interested in a more open `c`.

[...]

PR #1989, opened 2023-09-10
trying to fix mCrossTail

Logo121:
- Replaced some more `HVContrast` code to use `HSwToV`
- Some leftover code in `c.ptl` as I looked into https://github.com/be5invis/Iosevka/issues/1287#issuecomment-1708637055. No actual changes are made.
- Fixed a minor error with `upPointingHouse`
- Tried to fix `mCrossedTail` as proposed in #1882
	- Added a `_mid` parameter to a lot of `m`'s functions to control where the middle stem goes.
	- The 3 legs and the tail loop are spaced equally for now.

The result, tested with UltraCondensed and UltraExtended, Thin and Heavy:
Etoile
image
Slab
image

* It should build in all (permitted) weight and width settings now.
* However, it may make the glyph a bit awkward in ultra-extended + thin. I need some feedback on this:
    * Should I shift the middle leg back to the center on some variants only? or depending on stroke/width?
    * Should I just fix the leg to the center and let the tail overlap with it?
    * Or should I leave it as is?
* Other variants are also supported:
image
However, `m` breaks in some variants under ultracondensed. I think it exists before this PR, though I'm not sure.

### DIFF
--- a/changes/27.0.0.md
+++ b/changes/27.0.0.md
@@ -8,3 +8,4 @@
   - `upper-r`.`curly-open-motion-serifed` → `upper-r`.`curly-open-top-left-serifed`
   - `upper-r`.`standing-open-motion-serifed` → `upper-r`.`standing-open-top-left-serifed`
 * Add OpenType `zero` feature (#1966).
+* Fix broken geometry of `U+AB3A` under condensed width.
--- a/font-src/glyphs/letter/cyrillic/fita.ptl
+++ b/font-src/glyphs/letter/cyrillic/fita.ptl
@@ -9,8 +9,8 @@ glyph-block Letter-Cyrillic-Fita : begin
 	glyph-block-import Common-Derivatives
 
 	define FitaWave : XH / 16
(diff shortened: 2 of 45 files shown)


jmcwilliams403:
You've probably given it way too much clearance, as the loop doesn't have to stretch _with_ the second loop.

Logo121:
> the loop doesn't have to stretch _with_ the second arch.
you have a point.

My main goal was to make it not break on condensed versions, so I don't have a good idea on what to do on the stretched ones yet.

be5invis:
Please keep the space evenness of `mCrossedTail`. If the space is indeed highly limited, you can make the left side of the edge to slightly overlap with the `m` part's middle leg.

Logo121:
Looks like this for now, keeping the position of the curl while moving the middle leg back to the center:
image

As mentioned in #1782, it's really hard to prevent broken geometry while keeping the curl and the middle leg separate (even with the loop `fine` width thinner, there will be some corruption), so overlapping it is.

be5invis:
@Logo121 LGTM for this

Issue #1997, opened 2023-09-15
Latin x and cyrillic he (U+0425/U+0445) variants similar to new cyrl-zhe=cursive


pv4:
@be5invis @jmcwilliams403 

Is it possible please to create variants (nearly like on the following picture), similar to new cyrillic zhe variants?

![image](https://github.com/be5invis/Iosevka/assets/2502094/8704de26-daa7-4265-a734-2fc5b48f0b02)
![image](https://github.com/be5invis/Iosevka/assets/2502094/8019b3f4-085d-4e19-991f-d3f53c66172a)

This variant would be visually simplier than the current `x=cursive` and I personally would prefer it (currently I use standard/upright x variant in my italic builds).

In many fonts and in russian hand-writing cyrl-he/x/cyrl-zhe have simialr writing for legs:

1. Iosevka (straight)
![image](https://github.com/be5invis/Iosevka/assets/2502094/0586f1bf-41e0-47d1-be3f-215813a28254)
![image](https://github.com/be5invis/Iosevka/assets/2502094/e4e98423-fec4-4949-8f4a-7b93f3e59be7)

2. russian hand-writing
![image](https://github.com/be5invis/Iosevka/assets/2502094/53bd815e-441e-4d37-aefa-dd93e8a27163)

jmcwilliams403:
The cursive zhe variants were added for `ss12` and `ss15`.
IBM Plex Mono, the source for `ss15`, uses the existing cursive form of `x`/he, while Ubuntu Mono, the source for `ss12`, does not use any special italic form for it.

I have no authority on whether this variant is worth adding.
All I can say is to please provide examples of other monospace/programming fonts that do this.

pv4:
Since what I suggest is some sort of a "cursive" style, not a "simple" one, it's hard to find an existing monospace/programming font of the style in question. At least, I wasn't able to.

But I found some cursive fonts with the style (see below).

Before the images, I have to say that the style I'm asking for is something between the current "normal" and "cursive" styles. It's more cursive than the "normal" one, and it's more normal (and horizontally symmetric) than the current "cursive" style. Why I'm asking it now is because I think the existing cyrl-zhe=cursive legs can be reused somehow, and I think it would be fine to have cyrl-zhe/cyrl-he/x variants that look similar.

And now, the images, all are from https://www.fontspace.com/

https://www.fontspace.com/valley-forge-font-f102566
https://www.fontspace.com/auntha-clear-trial-font-f103009
[...]

PR #2007, opened 2023-09-26
cleanup cyrl-u variants under italics

jmcwilliams403:
Firstly, all italic/slab/slab-italic forms are corrected to follow `y` where applicable, except for `ss10`.

Notably, `ss10`'s Cyrillic Capital U no longer has to be held hostage by Cyrillic Lower U forcively inheriting `y`'s hard-cursive form;
Since Envy Code R does not originally have Cyrillic support, they have been disunified, but in a way that's still in harmony when italicized.

`Ууy`

All SSs under Sans Serif:
image
All SSs under Slab:
image


pv4:
@jmcwilliams403 Hello!

Sorry for off-topic, but could you have a look at #1997 please?

PR #2010, opened 2023-09-28
Flat 5 (or whatever that is called)

### LINKED ISSUE #1995: Variant for number five
Could create a more symmetrical Variant of 5?

Jetbrains Mono
jetbrains-mono_five2

Pragmata Pro
pragmata-pro_five2

Logo121:
Closes #1995.

* There are some inconsistency between using `local` or `define`, particularly for the recent `df : include : DivFrame ...` format.  
The `local` version seems to be more common (20X vs 4X), but I can't be so sure if this difference matters. I've made changes in files that has both versions (preferring the `local` one), which are:
	* `o.ptl`
	* `lower-n.ptl`
	* `zhe.ptl`
	* `lower-ae-oe.ptl`
	* `upper-h.ptl`
	I've checked if that breaks the affected glyphs but found nothing.
* Made `ᶃ` m-width in quasiproportional so that it looks less compressed. 
image
* Added `5` variant with flat middle stroke. Though I reaaallly have no idea what to call this variant.
image
	* Also made Zhuang Tone Five follow the variant selector (on the middle shape only) for now.
		* I've seen a few fonts that actually treat Tone Five a bit more differently, where only Tone Five has this flat shape, or even a higher joint (treating it like a derivative of `s`). Probably not an issue now.
* Clipped the serif for small Zhuang Tone Five since it overlaps with the bowl(?) in some settings.


Logo121:
Example for Five vs Tone Five (Charis SIL, Noto Sans Mono):
image
image (`5` is slightly arched)

Should I just use the flat version for tone 5?

jmcwilliams403:
I've always assumed `local` just means the variable is disposed of at the end of a `do` block, but that inside a `glyph-proc` block it's more of a convention thing — Or maybe it's important for garbage collection.
Be5invis knows the real answer.

Logo121:
> I've always assumed `local` just means the variable is disposed of at the end of a `do` block, but that inside a `glyph-proc` block it's more of a convention thing — Or maybe it's important for garbage collection. Be5invis knows the real answer.

I dunno, I found some cases where two `glyph-proc`s had very similar structure, but one uses `local` and the other uses `define` for their `df` declaration. That's what prompted me to at least fix the files with both of them coexisting.  
It didn't feel like they would've made a difference.

Isn't `glyph-proc` just a wrapped function block though? idk

jmcwilliams403:
> Isn't `glyph-proc` just a wrapped function block though? idk

I meant to imply that the distinction of scope between `local` and `define` when called within any wrapped function may possibly just be conventional at that point, but Be5invis knows the real answer. We're both only making broad assumptions.

be5invis:
Well `define` and `local` are almost the same tihng. The only difference is that `define` defines a constant. Their scoping rules are the same.
Ah I should write docs for PTL ...

PR #2022, opened 2023-10-03
add exe files to gitignore

jmcwilliams403:
This allows users/contributors to simply place the ttfautohint exe file into the Iosevka directory without modifying their global path variable on the OS level.


be5invis:
Maybe a better way to allow configuring external programs by using environment variables.

PR #2031, opened 2023-10-09
Use previous customizer preview as package sample

jmcwilliams403:
Now that they are properly synced up, it makes perfect sense to use the previous website version for both to show as many customized characters as possible


utility/generate-samples/templates/package-sample.mjs
```diff
@@ -3,9 +3,9 @@ import * as themes from "../themes/index.mjs";
 // prettier-ignore
 export const ssStrings = [
 	["ABC.DEF.GHI.JKL.MNO.PQRS.TUV.WXYZ", "abc.def.ghi.jkl.mno.pqrs.tuv.wxyz"],
-	["¢ſßðþ ΓΛΔαδιλμξ КУЗЯЖэльфязычникж", "float il1[]={1-2/3.4,5+6=7/8%90};"],
```
chelovechishko, line 6: > КУЗЯЖэльфязычникж

To note, these were almost meaningful words

PR #2032, opened 2023-10-09
add: variant: flower asterisk (aka eight-pointed)

chelovechishko:
For those, who wanted eight-pointed asterisk aka flower asterisk but doesn't know how, now you can:

- apply this changes to a building copy of yours
- use like: `asterisk = "flower-low"` in `private-build-plans.toml`

flower

NB, checked only 1 custom variant.
NB, entirely through github web interface.


chelovechishko:
And insta close.

PR #2046, opened 2023-10-15
Some miscellaneous CV corrections for several SS's, mostly involving new variants.

jmcwilliams403:
I originally started with polishing up Chi and Lambda variants for any applicable SS's which I forgot to mention earlier, and then later I incorporated several other smaller uncategorized fixes that would be wasteful to do in separate PR's.

Something to note is that Recursive Mono in particular actually uses an italic `x` that's closer to the new chancery style than the original cursive style:
image

All fixes are of course done with the original fonts in mind.


be5invis:
I **really really really really really really** hope that for the SS modifications you could try combine all your fixes in one PR instead of spreading them into multiple PRs. There are already #1871, #1944, #1946, #1961, #1972, #1985, #2016, #2023, #2024, #2043.

be5invis:
I made a tool, https://typeof.net/Iosevka/style-matcher, which may be useful for style matching... I hope.

PR #2075, opened 2023-11-05
Add orthographic italic form for Cyrillic Three-Legged Te (`U+1C85`) based on original Cyrillic Te.

jmcwilliams403:
`тᲅп`
Upright:
image
Italic:
image
Compared to Noto Serif:
Upright:
image
Italic:
image

### DIFF
--- a/changes/27.3.5.md
+++ b/changes/27.3.5.md
@@ -1,0 +1,1 @@
+* Add italic form of CYRILLIC SMALL LETTER THREE-LEGGED TE (`U+1C85`).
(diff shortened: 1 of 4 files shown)


Logo121:
I actually decided against adding italic forms for three-legged te and long-leg de in #1809.

Considering these variants are encoded as a "potentially contrastive variations" of the original characters, I wasn't sure if it was meaningful to then merge them back again in italic versions, even if Noto Serif does it.

Idk tho, maybe there are justifications on Noto's side that I happen to miss. There are unfortunately no other italic-aware(?) fonts that supports this character.

jmcwilliams403:
> Idk tho, maybe there are justifications on Noto's side that I happen to miss. There are unfortunately no other italic-aware(?) fonts that supports this character.

according to https://github.com/notofonts/latin-greek-cyrillic/issues/53 , justifications exist for a unified cursive italic form for specifically three-legged te (book example linked in thread) but not for long-legged de.

Issue #2018, opened 2023-10-01
Mordvin Cyrillic letters (`U+0514`..`U+0519`)


jmcwilliams403:
- The requested characters are ...
  - [ ] Latin
  - [x] Cyrillic
  - [ ] Greek
  - [ ] Punctuation
  - [ ] Diacritic / Mark
  - [ ] Symbol
- The requested characters are used in ...
  - [x] The alphabet / character set of a natural language.
  - [ ] The alphabet / character set of a constructed language.
  - [ ] The alphabet / character set of a transliteration or notation system.
  - [ ] Mathematical or other professional use in documents.
  - [ ] The symbol set of a programming language. For the languages that support using arbitrary symbols, please provide usages in well-known components or libraries.
  - [ ] The symbol set of a command-line program.
- [x] Some other monospace/programming fonts supported this character. Provide images below.

------

Adding these characters will complete the Cyrillic supplement block and finally close #983

`ԔԕԖԗԘԙ`

Consolas:
![image](https://github.com/be5invis/Iosevka/assets/37010132/f7e4549f-8145-4166-b18e-da4f5034f285)
Courier New:
![image](https://github.com/be5invis/Iosevka/assets/37010132/cf12a578-63c9-44cb-97d9-4d26939b6b2c)
Fira Mono:
![image](https://github.com/be5invis/Iosevka/assets/37010132/9adf936c-6cb5-4032-b393-f15e49123394)
Noto Sans Mono:
![image](https://github.com/be5invis/Iosevka/assets/37010132/19647c3e-3265-42c7-84e9-fb7edb9b8557)
Everson Mono:
![image](https://github.com/be5invis/Iosevka/assets/37010132/3be9a5c8-5f16-45e8-b87e-cb4ca2a6919e)
Fairfax HD:
![image](https://github.com/be5invis/Iosevka/assets/37010132/ac4984a3-c632-4aad-8c02-aa11a8969754)

Logo121:
oh i have been staring at these letters for a long time

how would it look like when `х` is set to the cursive variant though, I can't find any fonts supporting `ԕԗ` with that cursive `х`

be5invis:
@Logo121 Maybe...
`ԕ` -- just make the top-left part of the cursive stroke flat.
`ԗ` -- just keep the right half. May need to modify `XCursiveHalfShape` to support a "uniform" X-strand though.

jmcwilliams403:
[...]

PR #2088, opened 2023-11-12
Mordvin Letters

Logo121:
Very incomplete attempt on #2018. 

Known issues to be fixed:
* Changelog
* The diagonal tail of the cursive `x` is deformed, but I've no idea what caused it.
* Lha needs some better adjustment on the stroke width and space allocation, as it is known to overlap in other variants.
* Upper/Lower Rha's Ha part extrudes into the bowl of Er a bit in some settings.
* Lower Rha with cursive Ha still looks a bit off

Any reviews/suggestions on how to fix the above are welcome.


be5invis:
I think it is simpler to just merge it and let me fix the remaining bugs ...

PR #2104, opened 2023-11-28
Add double-serifed/serifed-tailed variants for `a`/`α`, add bottom{-right}-serifed variants for `d`/`q`/`U`.

### LINKED ISSUE #1949: Lower `a` single-storey eared variants with top serif
- [ ] At least two monospace/programming fonts, created by different designers, supported the requested variant. Provide images below.

------

Simulated below with Latin Alpha.
This would basically allow italic `a` to be in harmony with `g`/`q`/`d` in any configuration.
`ɑdq`
Courier new:
image

DejaVu Serif:
image

jmcwilliams403:
Closes #1949 .

`aæ𝖺ᶏɐꞛɑꭤ`
All `a` variants:
image
`α𝝰`
All alpha variants:
image

For the case of `U` and `d`, the bottom serif variants are because the serif may be used as an alternative to a tailed terminal.

`U𝖴ᴜꞾ`
All `U` variants:
image
`d𝖽ʤᶁɖȡɗᶑԃ`
All `d` variants:
image

For the case of `q`, having only a bottom serif is an extremely common variant I've seen around:

Liberation Serif:
image
IBM Plex Serif:
image
Roboto Serif:
image
Times New Roman:
image
Georgia:
image

`q𝗊ʠɋ`
all `q` variants:
image

### DIFF
--- a/params/variants.toml
+++ b/params/variants.toml
@@ -1541,9 +1541,10 @@ selectorAffix.scripta       = "singleStorey"
 [prime.a.variants-buildup.stages.double-storey-hook."*"]
 next = "bar"
 
-[prime.a.variants-buildup.stages.double-storey-hook.normal]
+[prime.a.variants-buildup.stages.double-storey-hook.hook-serifless]
 rank = 1
 keyAffix = ""
+descriptionAffix = "serifless hook"
 selectorAffix.a = ""
 selectorAffix."ae/a" = ""
 selectorAffix."a/sansSerif" = ""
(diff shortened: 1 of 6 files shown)


params/variants.toml
```diff
@@ -1546,20 +1552,19 @@ selectorAffix."a/turnABase" = ""
 selectorAffix."a/single"    = ""
 selectorAffix.scripta       = ""
 
-[prime.a.variants-buildup.stages.double-storey-hook.hook-serifed]
+[prime.a.variants-buildup.stages.double-storey-hook.corner-hooked]
```
be5invis, line 1555: Please do not rename this. The shape here in `a` does not shape like `r` (which has right angle).

be5invis (changes requested):

be5invis (approved):

PR #2106, opened 2023-11-29
Harmonize default italic slab variants for `a`, `g`, and `q`.

jmcwilliams403:
I wanted to do this as a separate PR to give you the ability to say yes or no to this part, separately from just having the `a` variants exist at all.
`agqagq`
Before:
image
After:
image


be5invis:
It is ok for this part.

PR #2109, opened 2023-12-03
Add two characters from Combining Diacritical Marks Supplement.

jmcwilliams403:
Not shown is the combining parentheses interaction with the circumflex tie which doesn't appear to function correctly in my firefox test environment, even for the existing tie characters, and notepad just clips the glyph off completely at the height of tie marks on Windows, so I'm unable to verify if there's an environment that shows the correct behavior.

```
V᷍VŴ
a᷻ a᷻᪻
```
Thin:
image
Regular:
image
Heavy:
image


Logo121:
Shouldn't the double circumflex be wider/as wide as double parenthesis?

Also, it's more accurate to interpret the deletion mark as a series of (3 or 4) stripes instead of a triple prime:
image
image

jmcwilliams403:
> Shouldn't the double circumflex be wider/as wide as double parenthesis?

Yes but the top got really thin, and this is how Microsoft fonts do it anyway. I'll leave it up to someone else if they want to figure out the Unicode-recommended glyph.

> Also, it's more accurate to interpret the deletion mark as a series of (3 or 4) stripes instead of a triple prime: 

Right now, it's close enough, as many other fonts draw it like this. The math for what the best stroke width would be was also really tedious to figure out. Again, I'll leave it up to someone else to improve upon this.

PR #2128, opened 2023-12-23
Remove tailless variants for `{latn|cyrl}/iota` for disambiguation.

jmcwilliams403:
All Iota variants:
`ιƖɩꙆꙇ`
image
All Tau variants:
`τꚌꚍ`
image
Additionally, for consistency, italic Serbian Cyrillic ghe defaults to flat-tailed when `i` is tailless.
All `i` variants:
`iг`
image

### DIFF
--- a/changes/28.0.2.md
+++ b/changes/28.0.2.md
@@ -2,7 +2,6 @@
   - LATIN CAPITAL LETTER AU (`U+A736`) (#1689).
   - LATIN SMALL LETTER AU (`U+A737`) (#1689).
 * Refine shape of Tshe and Cyrillic Capital Letter Te with Middle Hook (`U+A68A`) (#2123).
-* Remove tailless variants for Latin Iota (`U+0196`, `U+0269`) and Cyrillic Iota (`U+A646`, `U+A647`).
 * Remove bottom serif of Cyrillic Small Letter Ghe with Middle Hook (`U+0495`) under italics.
 * Make serif variants of Cyrillic Small Letter Tall Te (`U+1C84`) respond to italics.
 * Make terminal serif behavior of palatalized Komi consonants (`U+0502`...`U+0505`, `U+0508`...`U+050F`) more consistent with each other.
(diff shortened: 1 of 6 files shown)


changes/28.0.2.md
```diff
@@ -2,6 +2,7 @@
   - LATIN CAPITAL LETTER AU (`U+A736`) (#1689).
   - LATIN SMALL LETTER AU (`U+A737`) (#1689).
 * Refine shape of Tshe and Cyrillic Capital Letter Te with Middle Hook (`U+A68A`) (#2123).
+* Remove tailless variants for Latin Iota (`U+0196`, `U+0269`) and Cyrillic Iota (`U+A646`, `U+A647`).
```
be5invis, line 5: 28.0.2 is already out.

params/variants.toml
```diff
@@ -2328,7 +2328,7 @@ description = "Serifed `i`"
 selector.dotlessi = "serifed"
 selector."dotlessi/sansSerif" = "serifless"
 selector."dotlessi/compLigRight" = "serifed"
-selector."dotlessi/ital" = "tailedSerifed"
+selector."dotlessi/italic" = "serifedFlatTailed"
```
be5invis, line 2331: Change to `dotlessi/forceTailed`. Italicness is not a proper name here.

params/variants.toml
```diff
@@ -4772,72 +4772,84 @@ rank = 1
 description = "Greek lower Iota (`ι`) like a straight line"
 selector."grek/iota" = "serifless"
 selector."grek/iota/sansSerif" = "serifless"
+selector."grek/iota/tailed" = "flatTailed"
```
be5invis, line 4775: Could we just name it `selector."latn/iota"`? Maybe we could add a (currently identical) selector for `cyrl/iota`?

be5invis (changes requested):

params/variants.toml
jmcwilliams403: I don't think having two selectors is that important. I mainly just wanted to match the name of the tau selector, not create a neutral selector between latin and cyrillic.

PR #2138, opened 2023-12-31
Improve lower `r` for Etoile.

jmcwilliams403:
This completes parity of flat-hook/tail variants between Etoile and Aile.
```
```
Etoile Upright:
image
Etoile Italic:
image
Compared with Aile:
image


be5invis:
I think it is too narrow which is ugly. Maybe the non-compact one is better.

jmcwilliams403:
Regular:
image
Italic:
image

jmcwilliams403:
If neither are any good I can simply close this PR

PR #2149, opened 2024-01-03
Make Cyrillic Lower Zhe cursive under Slab Italic.

jmcwilliams403:
I've found that it's more-common-than-not for font families with both sans-serif and serif forms to use a cursive zhe under serif even when the sans-serif form uses an oblique zhe.
`ЖжХх`
Slab Upright:
image
Slab Italic:
image
Compared with DejaVu Serif:
image
Compared with Liberation Serif:
image
Compared with PT Serif:
image
Compared with IBM Plex Serif:
image
Compared with Noto Serif:
image
Compared with Times New Roman:
image
Compared with Georgia:
image


be5invis:
I'd like to hold this change because I dislike the current shape of cursive Zhe. It is somehow "underpowered" compared to cursive *x* and *z*.

PR #2173, opened 2024-01-17
Fix doc url links

nobk:


nobk:
So far I have found these two small problems

PR #2185, opened 2024-01-28
Some (legacy) terminal characters

Logo121:
Implements some characters from [this mapping table](https://www.unicode.org/L2/L2021/21235-terminals-supplement-sources.pdf) (+some related ones), where some characters from the legacy character sets are mapped to existing codepoints.

`⎹⎺⎻─⎼⎽⎸ ⛶□ ⧉⮺⮻⮼` (`─□` for reference)
image

For the scanlines, the only description I can find about them is [this mail message in unicode's archive](https://www.unicode.org/mail-arch/unicode-ml/y2014-m04/0041.html), which implies that all of the scanlines (and probably by extension the vertical box lines) are aligned with box-drawing characters, so they should have the same stroke width.

For the square four corners, they are realized as White Square with the sides cut, like in the mapping table and in FairfaxHD:
image

It seems like most fonts actually render them with rounded squares (as does Unicode's chart), but I suspect that is a consequence of a "rounded gothic" font being referenced, where many squared characters are also rendered as round squares:
https://ja.wikipedia.org/wiki/ARIB%E5%A4%96%E5%AD%97%E3%81%AE%E8%BF%BD%E5%8A%A0%E8%A8%98%E5%8F%B7%E4%B8%80%E8%A6%A7
image

As a reference here are some fonts designed for the ARIB codemap that uses a cornered square for this character:
- https://designpocket.jp/font/detail/19677 (https://www.fontucom.com/pdf/AFSARIBR-Map.pdf)
- https://culti1994.blogspot.com/2013/08/arib.html

For the overlapping (or joined) squares they just use the White/Black Medium Square with some translations.

Other remarks:
- `lower-r.ptl` changes are just minor replacements (using `HSwToV`), with no actual impacts to the glyphs. (hopefully).
- For `⛶⧉⮺⮻⮼` I have no idea where to place their implementations, so I just put them along with the "plain" squares. They can be put anywhere else without any problems.


jmcwilliams403:
I would also like to mention that all characters added here were implicated for unification back in the proposals for the original Symbols for Legacy Computing block in 2018, and funnily enough I was also planning to file character requests for them but you appear to have once again read my mind.

be5invis:
LGTM

PR #2201, opened 2024-02-09
Decorate White Arrows (#2063)

### LINKED ISSUE #2063: Decorated White Arrows (Keyboard Arrows) (`U+21EA`..`U+21F0`, `U+23CE`, `U+2BB8`)
- The requested characters are ...
  - [ ] Latin
  - [ ] Cyrillic
  - [ ] Greek
  - [ ] Punctuation
  - [ ] Diacritic / Mark
  - [x] Symbol
- The requested characters are used in ...
  - [ ] The alphabet / character set of a natural language.
  - [ ] The alphabet / character set of a constructed language.
  - [ ] The alphabet / character set of a transliteration or notation system.
  - [x] Mathematical or other professional use in documents.
  - [ ] The symbol set of a programming language. For the languages that support using arbitrary symbols, please provide usages in well-known components or libraries.
  - [ ] The symbol set of a command-line program.

------

These are all keyboard modifier symbols used in user interfaces or documentation. Adding these would make it easier to use them in comments, especially inside the sorts of box-drawing keyboard layout diagrams used in XKeyboardConfig, where fallback fonts would ruin the readability.

For example, `⇪` is caps lock or shift lock, `⮸` is caps lock when the former is used as shift lock, `⇮` is altgr, `⇯` is altgr latch, `⇰` is group switch, etc.

`⇪⇫⇬⇭⇮⇯⇰⏎⮸`
DejaVu Sans Mono:
image
Pragmata Pro:
image

Logo121:
Closes #2063.

Basically decomposing the arrow shapes into a few "heads" (or tails) and making several `doublearrow2`s from them.


jmcwilliams403:
I think the minimum gap size for the first and last arrows should be capped to a proportion of the total height using a min/max function, because they otherwise look quite different under the thin weight compared to the other weights.

Logo121:
the base and the disjointed rectangle thing are scaled with the arrow head length. They can be thinner if needed.

The gap itself is scaled with the stroke width, so maybe I'll change it to be based on the head length as well instead?

Other oddities are because of hinting/aliasing, I *think*.

Logo121:
image

Modified such that the white part of the rectangle is constant (instead of including the border). The gap also scales with that white part.

I suppose this is what you want?

jmcwilliams403:
> I suppose this is what you want?

LGTM

PR #2209, opened 2024-02-22
fix(hashFile): use path.basename to get file name

Zenvie:
Previously, the hashFile function was using file.name to get the file name, which resulted in undefined values in the output file. This commit fixes that by using path.basename(file.full) instead, which correctly extracts the file name from the full path.


Zenvie:
I apologize for the code I submitted earlier at [this pull request](https://github.com/be5invis/Iosevka/pull/2012), which did not fully fix the issue.

PR #2219, opened 2024-03-04
Add various dingbats and punctuation characters.

jmcwilliams403:
* For the 3D arrows, I've seen them in use as decorative bullets in Powerpoint presentations. They are handled similarly to other outlined half-filled shapes.
* For the omission brackets, it completes the set with raised square for biblical transcription.
  * the paraphrase brackets are used for different purposes but they look identical in the Unicode charts apart from being subscript.
* For the circled 18 with overlaid backslash, it happens to be one of two characters described as additional enclosed alphanumeric characters, the other of which (keycap 10) is not added here because it involves a keycap enclosure, whose glyph exists in Unicode as a combining enclosing character that we do not yet support.
  * here I have made it narrow by default unless the font is made quasi-proportional, as I feel it is more appropriate to treat it specially since it's named specially in Unicode ("no one under eighteen symbol") and it implies a restriction of access to a piece of media even if not necessarily copyright.
* Black heart is just the generic form of heavy black heart; Heavy black heart has an emoji form that is equivalent to a red heart, which is obviously out of scope. Under black and white text, both glyphs may look identical.
  * Maybe we can later distinguish them by making the heavy one fatter like how they are in the Unicode charts.
* The last one is just a raised left square bracket for the Unicode 16 proposed characters. A right square bracket is not proposed.
```
➢➣➤
⸋⸌x⸍⸜x⸝
🔞🖤
x🯌⁽
```
image


packages/font-glyphs/src/auto-build/composite.ptl
```diff
@@ -676,6 +676,10 @@ glyph-block AutoBuild-Enclosure : begin
 		foreach [j : range 36 till 50] : compositions.push : list (0x32B1 + j - 36) [digitGlyphNames j] WideWidth1
 		createCircledGlyphs 2 compositions
 
+	do "Double-digit backslash-circled"
```
be5invis, line 679: Consider remove.
If we are not adding keycaps then we should remove the circled 18 with overlaid backslash too.
(Actually, I do NOT want to add ANY MORE new Emoji into Iosevka.)

be5invis (changes requested):

Logo121:
@jmcwilliams403 Are you sure about `🯌` being just a superscript `[`? The [alpha name list](https://www.unicode.org/Public/draft/UCD/ucd/NamesList.txt) does not mention it being a superscript (unlike `⁽` and others), and FairfaxHD makes it smaller:
image

jmcwilliams403:
> @jmcwilliams403 Are you sure about `🯌` being just a superscript `[`? The [alpha name list](https://www.unicode.org/Public/draft/UCD/ucd/NamesList.txt) does not mention it being a superscript (unlike `⁽` and others), and FairfaxHD makes it smaller:

Well the name says "small" so it could stand to be changed if you want.

jmcwilliams403:
@Logo121 fixed in #2222 :
`x🯌ˣ⁽ᴴ`
image

PR #2249, opened 2024-03-21
Remove untagged variant selector for Cyrillic Capital Ef from #2155 as it was never a good idea.

jmcwilliams403:
The variants are rare and the split bowl is extremely sensitive to breaking after having adjusted `diversityM`.

As far as the default configurations are concerned, the font will present as unchanged to the end-user.


be5invis:
I'd like keeping the selector for Serifless and Serifed.

PR #2253, opened 2024-03-22
Miscellanious last-minute fixes and optimizations.

jmcwilliams403:
* Slightly optimize leaning anchor for gha.
* Cleanup leftover unneeded parentheses anchor code for circumflex tie from when it was just a centered single mark.
* Fix discontinuity for `7` variants for `ss08` between sans and slab.
* Require at least three hyphens to trigger hyphen chain for C-likes for better disambiguation between a decrement operator token and just a continuous horizontal bar.

If possible, it would be nice if this could be merged without squashing, but it's not super important.


be5invis:
I'd like to use squashing for a cleaner history in `dev`. Sorry for that.

PR #2272, opened 2024-04-02
Optimize the build speed by producing simpler arcs when converting spiro to outline

be5invis:


jmcwilliams403:
It's possible either this PR or #2270 broke something in o-shapes and dots:
image

be5invis:
@jmcwilliams403 Run `npm i` to update deps (`spiro` in particular -- I fixed a bug causing this earlier)

PR #2279, opened 2024-04-04
Three modifier letter additions, slightly adjust Latin-1 Macron.

jmcwilliams403:
First, a demonstration of Latin-1 Macron next to Modifier Letter macron:
`x¯ˉ`
image
The Latin-1 version is adjusted to extend to the side bearings of `n` for disambiguation, which is a similar distinction to how Ascii Grave and Latin-1 acute are made slightly larger than a true diacritical grave/acute.

Secondly, three additional modifier characters are added:
`꜠꜡ꜚ`
image

I modeled them after Liberation Mono:
image


jmcwilliams403:
Closing because the stress and low tones should be derived from different glyphs than shelves. I'll reopen with a reduced PR without them for now.

Issue #2271, opened 2024-04-02
Connecting orthogonal and diagonal box-drawing glyphs


Logo121:
Continuation of #2197.

This problem exists before the introduction of the Unicode 16.0 box-drawing diagonals, where the "diamond" diagonals do not connect with the orthogonals ones `🮮┼`
![image](https://github.com/be5invis/Iosevka/assets/21302803/e697125a-a83a-4370-b7b6-a6139f0fb3f5)

To my understanding the "cell-masking" done in those diagonals are to remove the gaps that appear when 2 (non-parallel) lines join, which I followed for the diagonal lines in 16.0. However, in the schematic symbols section we have these:
![image](https://github.com/be5invis/Iosevka/assets/21302803/c44c3291-fdeb-4e08-8262-18918719b498)
Which kinda assumes connection with diagonals and orthogonal lines (at least for these 4 characters. I can't really find examples of circuit diagrams with those microcomputers so I can't tell if diagonal/orthogonal connections are needed in general). Right now connecting them results in something like this:
![image](https://github.com/be5invis/Iosevka/assets/21302803/96b44634-71b3-43c5-8fef-2aec7bf91c6d)

The potential solutions are:
1. Make all lines unclipped, and only deal with cases where gaps appear in a single character (like `🯘🯙🯚🯛`)
	- For the other cases like `🮠🮡 🮢🮣`, we just ignore the gap problem, or try to reduce it somehow like using a rounded cap?
	- Essentially what FairfaxHD does.
2. Modify the width of the diagonal lines so that they connect to orthogonal lines directly.
	- This is cleaner but would probably result in lines varying quite a bit in thickness. Also requires some calculations for every kind of slopes.
1. Deal with the schematic symbols directly and assume no diagonal-orthogonal connection should happen otherwise.
	- This means making `𜰉𜰊` unclipped unlike the other diagonal lines, or making the vertical connection part of `𜰐𜰑` a spline like in FairfaxHD: ![image](https://github.com/be5invis/Iosevka/assets/21302803/fa392ad4-6b27-4f78-8bed-a8129b276db3)
	- Note that the spline form is not used in the Fullwidth equivalent (KreativeSquare), which is the Unicode reference glyph.

... or a mixture of the above.

While we're at it we might also want to disable HVContrast for all box-drawing glyphs, if that makes the situation better.

Affected glyphs:
[...]

Issue #2189, opened 2024-02-02
Symbols for Legacy Computing Supplement & other Unicode 16 semigraphics


be5invis:
- The requested characters are ...
  - [ ] Latin
  - [ ] Cyrillic
  - [ ] Greek
  - [ ] Punctuation
  - [ ] Diacritic / Mark
  - [x] Symbol
- The requested characters are used in ...
  - [ ] The alphabet / character set of a natural language.
  - [ ] The alphabet / character set of a constructed language.
  - [ ] The alphabet / character set of a transliteration or notation system.
  - [ ] Mathematical or other professional use in documents.
  - [ ] The symbol set of a programming language. For the languages that support using arbitrary symbols, please provide usages in well-known components or libraries.
  - [x] The symbol set of a command-line program.
- [ ] Some other monospace/programming fonts supported this character. Provide images below.

------

There are a lot of symbols under the new **Symbols for Legacy Computing Supplement** block. Some of them are very tricky to make (like the outlined letters -- we **MAY** need to build an automated outlining algorithm to generate these).

L2: https://www.unicode.org/L2/L2021/21235-terminals-supplement.pdf

Logo121:
A quick progress report: After #2293, these symbols are left to be implemented, categorized by their shape:

### Dots
- [ ] Speckles `𜱆𜱇`
- [ ] Explosion `𜷽𜷾𜷿`
The laziest implementation is probably to implement them like mosaics (and hardcoding each dots), but there might be better ways.

### Terminal shapes
- [ ] Non-round voided cross `🯋`
- [ ] Black Small Chevron `🯍`
Both have a close equivalent `🯀🯊` so implementations depend on them. Not sure how.

### Symbols with curves
- [ ] Curved Gates `𜰕𜰖`
- [ ] Coil `𜰋𜰌𜰍`

### Game sprites without curves
- [ ] Rocket `𜱖𜱗𜱘𜱙`
- [ ] Moon Lander `𜷵`
- [ ] Rocket Booster? `𜲠𜲡` (Looks like a teletext mosaic to me)
- [ ] Roller Coaster? `𜲢𜲣`

### Game sprites with potentially simple curves
- [ ] Rifle `𜱥𜱦𜱧𜱨`
- [ ] Bassinet `𜱈𜱉`
- [ ] Tank `𜲞𜲟`
- [ ] Bomb `𜱠𜱡𜱢𜱣`
- [ ] UFO `𜲤𜲥`

[...]

PR #2293, opened 2024-04-14
Some more Legacy Computing Symbols (pt 8, mostly split graphic)

Logo121:
Part of #2189

Changes to schematic symbols
```
𜰊𜰉┼🮮 𜰑┼𜰉𜰉𜰐𜰑
┼𜰉𜰊𜱃 ┼𜰐𜱃𜱃𜰐𜰊
```
image

Snake Head `𜱰𜱱𜱲𜱳𜱴𜱵𜱶𜱷`
image

Faces
```
😐😶🙁🙂🙃𜱯☹☺☻
𜲦𜲧𜲨𜲩𜲦𜲧𜲨𜲩
𜲪𜲫𜲬𜲭𜲮𜲯𜲰𜲱
```
image

Television + Microcomputer
```
𜲲𜲳𜲶𜲷
𜲴𜲵𜲸𜲹
```
image

Robot + Forward runner
```
𜱚𜱛𜱜🅯𜷰𜷰𜷰𜷰𜷶𜷷
𜱝𜱞𜱟🅯𜷱𜷲𜷳𜷴𜷴𜷴
```

image

</details>

---

Covered Ranges:
- `U+1CC70` - `U+1CC77` (Snake heads)
	- There are no "snake body" equivalent apparently, even in the original char map.
	- I just assumed they would use the heavy box-drawing glyphs as bodies, though in most case this makes their body very thin compared to its head.
	- I'm keeping it as is unless there's a better solution.
- `U+1CCA6` - `U+1CCB1` (Big faces)
	- Refactors the existing face a bit so that it can be extended to the 2x2 versions, hopefully keeping the original ones intact.
	- Probably has a lot of room for improvement.
- `U+1CCB2` - `U+1CCB5` (Big TV)
	- I kinda want to make the TV screen sides a bit more curved, but I've no idea how at all.
	- Currently the NWID version is so thin that it looks more like a Wii. It is possible to rearrange the buttons to the bottom to make the proportion slightly better but I'm not sure if that will be a good idea.
- `U+1CCB6` - `U+1CCB9` (Big Microcomputer)
	- Keys are hard-coded according to the reference glyph.
	- Again it's possible to rearrange the keys to make them less squished (e.g. numpad-style?) but I'm not sure if I should do it.
- `U+1CC5A` - `U+1CC5F`, `U+1CDF6` - `U+1CDF7` (Robot characters)
	- Probably a lot to be refined here. In particular I have no idea what the negative space inside the robot body is supposed to be.
- `U+1CDF0` - `U+1CDF4` (Front-facing runner)
	- Extended CC Human Figure's design for the front-facing runner.

Others:
- Attempt on resolving #2271, using the easiest way (imo)
	- Resistors `𜰉𜰊` now have their terminals clipped perpendicularly(?) to the cell border only.
		- This makes connections to both itself and to an orthogonal line seamless at least
		- Anything else will fail in one way or another.
	- Transistors `𜰐𜰑` now have their terminals clipped on both the cell border and right side of the vertical line.
		- This will leave a small gap for connections to both itself or to a vertical line, but is the only compromise I can think of using a simple clip
	- There are probably no "optimal" solutions here, but there are 2 other potential ways, both requiring (or work better with) removing HVContrast across all box-drawing characters:
		- Calculating the precise points of intersection for the dispiro borders of diagonal and orthogonal lines, then extending them up to that point
		- Adding a rounded cap to all line terminals, so the connection gaps are in the worst case rounded corners


PhMajerus:
I think you inverted the snake open and closed mouth versions:

image

image

About the body pieces, I believe they reused the energy wave characters:
image

This guess is based on the fact that those snake heads come form the Sharp MZ European alternate character set:
image

So the screen must have only used characters from that set, and the energy waves are the only ones that I can see that could provide the 4 directions and could look like scales when using the same color as the heads.
Something like one of the following:
image
image

However I couldn't find any screenshot to support that theory, and considering that these characters were only available on the European variant means many games used other characters to draw snakes.
But they may also have used the medium or large squares:
image
image

Game developers were pretty much free to use whatever character they wanted, as long as it was from that set.

PR #2294, opened 2024-04-14
Refine tails of Schwa and Reversed Lower E under italics

jmcwilliams403:
Basically, because they all bypass the swash code (including incidentally reversed e, even before the schwa serif code, simply due to having to redraw it mirrored), they shouldn't need to correct the tail metrics externally because they just end up hypercorrecting for nothing, causing it to curl too far inward, which may be problematic in the upper extremes.

Additionally, my `suppressSwash` parameter is renamed to `noSwash` for brevity.

Also, since Reversed e would now assume upright/oblique hook metrics all the time (because the hook being on the left means it can't have a swash), it shouldn't need to call the entire `xTerminalR` function in its entirety, especially since it has to add code to account for it being on the left side. Replacing it with an exact expression means `SmallESerifedTerminalShape` would be the only place left that calls the function earnestly, so I just combined `xTerminalR` into into that function inline for simplicity, including an extra `&& !schwaTail` clause that goes with `noSwash`, which Reversed e would just send `true` to at all times if it were still a function.

```
ЭƏЗэəз
Cce Ↄↄɘ Ɔɔǝ
```
sans upright (for comparison; effectively unchanged):
image
sans italic:
image
slab upright (effectively unchanged):
image
slab italic:
image

### DIFF
--- a/changes/29.2.1.md
+++ b/changes/29.2.1.md
@@ -1,4 +1,5 @@
 * Refine shape of `U+1CC09`, `U+1CC0A`, `U+1CC10` and `U+1CC11`.
+* Refine terminal shape of Schwa and Reversed Lower E (`U+018F`, `U+0258`..`U+025A`, `U+04D8`..`U+04D9`) under italics.
 * Add characters:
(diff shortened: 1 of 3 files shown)


jmcwilliams403:
don't merge yet, fixing inward slabs under heavy

jmcwilliams403:
While slightly improved, it still suffers from a deeper issue that also affects Cyrillic Ze:
image

jmcwilliams403:
The solution was to use the first part of the `Math.min` function for the serifed variants and the second part for everything else.
Heavy Italic Slab:
image

Issue #2289, opened 2024-04-12
Erase to the Left, Erase to the Right, and X in a Rectangular Box (`U+2326`‥`U+2327`, `U+232B`) (Keyboard Symbols).


jmcwilliams403:
- The requested characters are ...
  - [ ] Latin
  - [ ] Cyrillic
  - [ ] Greek
  - [ ] Punctuation
  - [ ] Diacritic / Mark
  - [x] Symbol
- The requested characters are used in ...
  - [ ] The alphabet / character set of a natural language.
  - [ ] The alphabet / character set of a constructed language.
  - [ ] The alphabet / character set of a transliteration or notation system.
  - [x] Mathematical or other professional use in documents.
  - [ ] The symbol set of a programming language. For the languages that support using arbitrary symbols, please provide usages in well-known components or libraries.
  - [x] The symbol set of a command-line program.
- [x] Some other monospace/programming fonts supported this character. Provide images below.

------
These are keyboard symbols used in UI and also comments/documentation of keyboard libraries such as [xkeyboard-config](https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config):
![image](https://github.com/be5invis/Iosevka/assets/37010132/cd2c4c4f-4dd6-48fa-bfe0-5a628d3e0081)
`⌫` is back delete (modern backspace), `⌦` is forward delete (key literally labeled "delete" on PC's), and `⌧` is clear (keypad 5 when numlock is off, or numlock itself on Mac).
`⌦⌧⌫`
DejaVu Sans Mono (My favorite, and also the closest to the Unicode charts):
![image](https://github.com/be5invis/Iosevka/assets/37010132/ded0dbb8-912e-4e9e-8ea3-d7b86fcad748)
Pragmata Pro:
![image](https://github.com/be5invis/Iosevka/assets/37010132/b6f9118d-1348-4237-87f7-69aae2da89bb)
Jetbrains mono (does not have clear symbol):
![image](https://github.com/be5invis/Iosevka/assets/37010132/5575e57f-97cd-4a96-86dd-1f6ef39f8eb0)

Issue #2192, opened 2024-02-03
The other Legacy Computing characters unified to existing codepoints


Logo121:
- The requested characters are ...
  - [ ] Latin
  - [ ] Cyrillic
  - [ ] Greek
  - [ ] Punctuation
  - [ ] Diacritic / Mark
  - [x] Symbol
- The requested characters are used in ...
  - [ ] The alphabet / character set of a natural language.
  - [ ] The alphabet / character set of a constructed language.
  - [ ] The alphabet / character set of a transliteration or notation system.
  - [ ] Mathematical or other professional use in documents.
  - [ ] The symbol set of a programming language. For the languages that support using arbitrary symbols, please provide usages in well-known components or libraries.
  - [x] The symbol set of a command-line program.
- [x] Some other monospace/programming fonts supported this character. Provide images below.
(Partially)

------

Here is a list of some more characters in the Legacy Computing charsets that are unified with existing, non-emoji codepoints.

Note that I may have missed some, considering my methodology for searching these character is:
1. Grep all the mapping `TXT` files attached to the proposal for codepoint-like substrings (hexadecimals with more than 2 digits)
2. Sort and format all of them into a readable format
3. Filter characters (by hand) that are:
    - Out of scope for Iosevka (mostly arabic, japanese and korean characters)
    - Unassigned (or assigned to the new codepoints)
4. Eyeball characters that aren't rendered in Iosevka and pick those characters out

The following I can confirm that are definitely mapped from those charsets because they are in the [sources table](https://www.unicode.org/L2/L2021/21235-terminals-supplement-sources.pdf):
```
- ☂: UMBRELLA (U+2602)
- ☙: REVERSED ROTATED FLORAL HEART BULLET (U+2619)
- ♟: BLACK CHESS PAWN (U+265F)
- ⛵: SAILBOAT (U+26F5)
- ✈: AIRPLANE (U+2708)
- ⧋: TRIANGLE WITH UNDERBAR (U+29CB)
```
(black chess pawn is part of #1418, which is eventually needed for the big versions in the supplement block anyway)

These are the extra ones I found:
```
- ₢: CRUZEIRO SIGN (U+20A2)
- ℞: PRESCRIPTION TAKE (U+211E)
- ⌁: ELECTRIC ARROW (U+2301)
- ⌇: WAVY LINE (U+2307)
- ⌛: HOURGLASS (U+231B)
- ⍻: NOT CHECK MARK (U+237B)
[...]

PR #2297, opened 2024-04-16
Collection of mostly technical symbols

Logo121:
Covers some of #2192.

---

<details>
<summary>Preview</summary>

Dentistry symbols (+some other symbols for aligning):
```
⎾⏁⏄⏇⏉⏋
⎸⏀⏃⏆❘⎹
⎿⏂⏅⏈⏊⏌
```

image

The rest: `⍻ ⎍⍾ ⎅⍿‖| ⏢⏤⏥⌒⌓⌔ ⏛🮯⏚ ⌫⌧⌦☒⮽⮾⮿⮹`
image

</details>

---

- `U+23C0` - `U+23CA` (Dentistry symbols)
	- Explanations on how these symbols work can be found by this blog post from the `⍼`-man himself: https://ionathan.ch/2022/04/07/dentistry-symbols.html
	- But basically, these are another set of symbols that are intended to be used with box drawings. Most of them are compositions of existing glyphs (except the waves)
	- Has a minor problem with the triangles and circles being affected by italic (not an issue by itself but is inconsistent with the wave shape). Not sure if there's a way to force them to be upright without creating more glyphs.
- `U+23BE` - `U+23BF`, `U+23CB` - `U+23CC` (Dentistry symbols for quadrants)
	- These are unfortunately more annoying to handle considering its possible forms:
		- Reference glyphs from the original proposal shows that these are vertical lines with top/bottom left/right horizontal line extensions, with the vertical lines usually centered. (like Kreative Square's version)
		- However, the Legacy Computing set encodes an almost identical set of "vertical + horizontal extension" lines: `𜸖𜸗𜸘𜸙`, while mapping the dentistry ones to "border-touching" box-drawing-like characters in the ISO 2407 standard.
		- Non-mosaic forms (leaving some margin from the border) for both representations above exist, considering they can both be used in non-mosaic contexts.
	- For now, they are just implemented as the 2nd variant (touching the cell borders), but kept at NWID by default (the rest are WWID) so that they look more uniform.
- `U+237F` (Vertical line with dot)
	- Simple composition.
- `U+2385` (White square with vertical line)
	- Simple composition. Assuming the gap is not necessary (following PragmataPro)
- `U+2312` - `U+2314` (semi/quarter white circles)
	- Simple masking and composition. Made for the bell symbol.
- `U+238D` (Monostable), `U+237E` (Bell Symbol)
	- Bell is implemented as a composition of Monostable and Segment.
- `U+237B` (Not check mark)
	- The extra stroke is adapted from the cross mark instead of a horizontal line because this looks way more natural. Also Amstrad CPC uses this variant.
- `U+23E4` `U+23E5` `U+23E2`
	- These are plain shapes. Implemented as such (including straightness which is a straight line but following GeometricDim here).
	- Flatness symbol is simply a parallelogram as far as I can tell. Aliased to whiteParallelogram.
- `U+2326` `U+2327` `U+232B` (Delete key shapes #2289)
	- Implemented as an extension of ballot box with cross.
- `U+2BB9` `U+2BBD` `U+2BBE` `U+2BBF` (Other ballotbox-like shapes)
	- Created along with the above.
	- I would like to mention that "Ballot Light-X" and "Circle Bold-X" are weird mappings, considering in the original (and reference) glyphs:
		- Ballot Light-X actually corresponds (in boldness) to Circle X, and Ballot X to Circle Bold-X
		- The Ballot Box and Circles vary their boldness along with the X
	- For now, they are implemented according to their name because there are no "lightWhiteCircle"s. It's unlikely any of the implementations will break if the glyph weights are swapped, so feel free to change them if needed.
- `U+23DA` `U+23DB` (Earth Ground, Fuse)
	- It's not explicitly mentioned whether they would connect to box-drawing glyphs like the new schematics or not, but I guess it wouldn't hurt.
	- The fuse has the same italic problem with the dentistry symbols.

Other:
- Left a commented implementation of `♾`. This is essentially a circled infinity but it fails to build due to some ordering issues.


jmcwilliams403:
FYI, the way you setup `⍿` makes it not respond to dot variants properly (nor forced-upright bar variants for that matter, but that may be less important).

jmcwilliams403:
Also you forgot to put the characters that close #2289 in the changelog (I closed it anyway though).

PR #2311, opened 2024-04-23
Finish work started by #2310 including important fixes.

jmcwilliams403:
* Fix soft dotting of all `i`/`j`-derived characters.
* Use a better name for the tittle variant selector.
* Finish tittle composition for the rest of the `i`/`j`-derived letters.
* Make `ss10` use square tittle.
* Add Turkish/Azerbaijani locales which use original diacritical dot behavior for `i` in particular.

```
<span style="font-family:'Iosevka Custom';font-feature-settings:'ss10';">
<p lang="tr">
Pijamalı hasta yağız şoföre çabucak güvendi.
</p>
<p lang="az">
Zəfər, jaketini də papağını da götür, bu axşam hava çox soyuq olacaq.
</p>
</span>
```
image


jmcwilliams403:
Reasoning for the naming of `tittle` and also why these locales do not need to affect `j`:
[ref.](https://en.wikipedia.org/wiki/Turkish_alphabet#Distinctive_features)
> Dotted and dotless I are separate letters, each with its own uppercase and lowercase forms. The lowercase form of I is ı, and the lowercase form of İ is i. (In the original law establishing the alphabet, the dotted İ came before the undotted I; now their places are reversed.) The letter J, however, uses a tittle in the same way English does, with a dotted lowercase version, and a dotless uppercase version.

PR #2315, opened 2024-04-27
fix git ignore macos ds store file

aprosail:
The desktop service store file (`.DS_store`) on macOS will usually been created during the build process. Ignore such file makes the code base tidy.


aprosail:
Sorry, I don't know the contribution rules. I'll close such PR and commit again into the dev branch.

Issue #2263, opened 2024-03-28
Remaining letters with palatal/retroflex hooks in Latin Extended G (`U+1DF12`‥`U+1DF1D`).


jmcwilliams403:
- The requested characters are ...
  - [x] Latin
  - [ ] Cyrillic
  - [ ] Greek
  - [ ] Punctuation
  - [ ] Diacritic / Mark
  - [ ] Symbol
- The requested characters are used in ...
  - [ ] The alphabet / character set of a natural language.
  - [ ] The alphabet / character set of a constructed language.
  - [x] The alphabet / character set of a transliteration or notation system.
  - [ ] Mathematical or other professional use in documents.
  - [ ] The symbol set of a programming language. For the languages that support using arbitrary symbols, please provide usages in well-known components or libraries.
  - [ ] The symbol set of a command-line program.
- [x] Some other monospace/programming fonts supported this character. Provide images below.

------

I should have probably asked for this before #2260 .

![image](https://github.com/be5invis/Iosevka/assets/37010132/c61275b5-7aaf-493f-9b78-b00fbf02bbc8)

`𝼒 𝼓 𝼔 𝼕 𝼖 𝼗 𝼘 𝼙 𝼚 𝼛 𝼜 𝼝`

Noto Sans Mono:
![image](https://github.com/be5invis/Iosevka/assets/37010132/0ffa99a7-b6fd-44e7-b7f1-aea320c9fc98)

Fairfax HD:
![image](https://github.com/be5invis/Iosevka/assets/37010132/b4c4deb6-7fd0-44c5-b08f-5d6924144307)

PR #2328, opened 2024-05-06
Add five letters with palatal hook.

jmcwilliams403:
Continuation of #2327 .
These are all the (non-affricate) palatal hook letters from #2263 .
`𝼓𝼔𝼕𝼖𝼘`
Sans Regular:
image
Sans Italic:
image
Slab Regular:
image
Slab Italic:
image


jmcwilliams403:
By the way something I can't fix is the eng's original EngHook on heavy:
image
I think it's something to do with the EngHook function not applying sw to its VerticalHook implementation, outside of x/y positioning.

jmcwilliams403:
Hopefully this fixes it:
image

PR #2381, opened 2024-06-17
Draft: Add bottom-serifed variants to Cyrillic Ze.

jmcwilliams403:
(Marking as draft because I cannot figure out how to get the `nonBreakingVariantAdditionPriority` argument to shift the new variants to the end of the stack instead of just permanently assigning them an index at the end.)

Basically, I took a closer look at some of the fonts I talked about in #2380 and saw that they tended to also put bottom serifs on Cyrillic Ze as well.
Mind you, these variants are mostly used under italics.

DejaVu Serif Italic:
image
Palatino Linotype Italic:
image

All Cyrillic Ze variants:
`ƐЗꚄꚈѮȜ`
image
`ɛᶓзᶔꚅꚉѯȝ`
image


be5invis:
Your current config should already work -- or you can bump version to 31.

PR #2409, opened 2024-07-07
Make `revs`-derived letters follow `cyrl/{Ze|ze}`.

jmcwilliams403:
Basically, because the top serif of rev-S is in the same "orbit" as the top serif of Cyrillic Ze, and because Cyrillic Ze's CV feature is sophisticated enough to know when it's appropriate to populate Latin Epsilon's bottom serif, which itself is in the same "orbit" as rev-S's bottom serif, this means that it may actually be more consistent to make rev-S's serifs follow Cyrillic Ze instead of unreversed S for rotational symmetry/balancing purposes.

On that note, unreversed/ASCII `S`/`s` is unchanged.

```
ꙄꙅГгӶӷӺӻ
ЗꙄСзꙅс
```

Slab upright:
image
Slab Italic:
image
`ЅѕГгӶӷӺӻ`
Compared to Vollkorn under Bulgarian:
image
Compared to Source Serif 4 under Bulgarian:
image

All Cyrillic Capital Ze variants:
`ƐЗꚄꚈѮȜꙄ`
image
All Cyrillic Lower Ze variants:
`ɛᶓзᶔꚅꚉѯȝꙅгғӷӻ`
image

### DIFF
--- a/changes/30.3.3.md
+++ b/changes/30.3.3.md
@@ -1,0 +1,1 @@
+* Make Reversed S (`U+01A7`..`U+01A8`, `U+A644`..`U+A645`) and Italic Cyrillic Lower Ghe (`U+0433`, `U+0453`, `U+0493`, `U+04F7`, `U+04FB`) follow variants of Cyrillic Ze (`cv69`..`cv70`).
--- a/packages/font-glyphs/src/letter/latin/s.ptl
+++ b/packages/font-glyphs/src/letter/latin/s.ptl
@@ -337,8 +337,8 @@ glyph-block Letter-Latin-S : begin
 	select-variant 's' 's'
 	link-reduced-variant 's/sansSerif' 's' MathSansSerif
 	select-variant 'smcpS' 0xA731 (follow -- 'S')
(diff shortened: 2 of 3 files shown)


be5invis:
You should make a new selector rather than make it follow Cyrillic Ze, as reverse S is very different from Ze shapes.

changes/30.3.3.md
```diff
@@ -0,0 +1 @@
+* Make Reversed S (`U+01A7`..`U+01A8`, `U+A644`..`U+A645`) and Italic Cyrillic Lower Ghe (`U+0433`, `U+0453`, `U+0493`, `U+04F7`, `U+04FB`) follow variants of Cyrillic Ze (`cv69`..`cv70`).
```
be5invis, line 1: You should make a new selector rather than make it follow Cyrillic Ze, as reverse S is very different from Ze shapes.

be5invis (changes requested):

jmcwilliams403:
Nevermind this PR for now. Maybe someday I'll write a different PR for a selector for Cyrillic Lower Ghe in particular, but not right now; It's not that important for the time being.

PR #2412, opened 2024-07-11
Return `param/private-parameters.toml` to git ignore.

jmcwilliams403:
As demonstrated in the example given in #2186 , having the ability to override the default parameters toml is currently one of the only ways to compile a custom font with different `diversityM` etc. values. The line was removed a few months ago in 2a19d92 but the feature still exists otherwise.

Perhaps it would be more constructive to mention it in the documentation than to silently remove it.


be5invis:
I think we could simply remove the support of `private-parameters.toml` though.

PR #2417, opened 2024-07-15
Fix docs regarding webfonts

viperML:
The interface expects the webfonts names in all uppercase, in https://github.com/be5invis/Iosevka/blob/main/tools/misc/src/make-webfont-css.mjs


viperML:
oops

viperML:
=> #2418

Issue #1835, opened 2023-07-08
Playing Cards (`U+1F0A0`..`U+1F0F5`)


jmcwilliams403:
- The requested character is...
  - [ ] Latin
  - [ ] Cyrillic
  - [ ] Greek
  - [ ] Punctuation
  - [x] Symbol
- [x] Some other monospace/programming fonts supported this character. Provide images below.

------

Pragmata Pro, representing Knight as C for Cavalier, which is the french word for Knight (I prefer using this label for this card):
![image](https://github.com/be5invis/Iosevka/assets/37010132/76b1be29-12a7-411d-bd5a-7b9621189736)

Fairfax HD, representing Knight as N for K[n]ight, and also representing the 0 thru 21 trump cards as Roman Numerals (this is actually how these trump cards are labeled in real life to contrast with the normal arabic 1 or 2 thru 10):
![image](https://github.com/be5invis/Iosevka/assets/37010132/2b816df0-dab2-4547-ae56-ceb298bc81d7)

Additionally, if the complete block is supported, it's common for Ace to be labeled as 1 instead of A in the complete real-life French Tarot deck which this block represents, but it's up to you if you care about going that far.

Logo121:
related: #495

though I guess another set of characters (#1418) in that issue got put into backlog too so

jmcwilliams403:
I made a rough mockup using the boxed enclosure functions.
Not sure if I'm all that happy with it yet but here's what it looks like so far:
![image](https://github.com/be5invis/Iosevka/assets/37010132/fdfd47a6-65b9-4f8a-ae5e-4492377ad61c)

PR #2419, opened 2024-07-15
Draft: Add playing cards (`U+1F0A0`‥`U+1F0F5`) (#1835).

jmcwilliams403:
Marking as a draft because this execution both feels and looks kind of hacked together in its current state. I'd really like to improve it somehow.

The digits currently aren't tailored to fill the vertical real estate they have access to, so two-row cards still use the same inner height as enclosed alphanumerics, which is one inner cap-height character.

I did go ahead and make a non-descending `Q` glyph for Queen (along with the existing one for `J` for Jack), which looks more visually consistent with the rest of the ranks, and it would probably look even better if the inner text filled the vertical space more.

I also didn't know what to put on the playing card back character (if anything) so I made it an empty negative boxed character. Maybe it could potentially have some sort of ornamental design on the back, but that's up to you.

Regarding the inclusion of the Red Joker:

Believe it or not, the "red" joker here isn't the one that actually has a registered emoji form (the black joker is), but due to the unfortunate name, many non-emoji fonts tend to neglect the red joker and only support the black and white.

Here, I have created a vertical shaded star for the red joker based on the black and white example glyphs for colored geometric shapes shown in the Unicode charts.

```
🂠🂡🂢🂣🂤🂥🂦🂧🂨🂩🂪🂫🂬🂭🂮
 🂱🂲🂳🂴🂵🂶🂷🂸🂹🂺🂻🂼🂽🂾🂿
 🃁🃂🃃🃄🃅🃆🃇🃈🃉🃊🃋🃌🃍🃎🃏
 🃑🃒🃓🃔🃕🃖🃗🃘🃙🃚🃛🃜🃝🃞🃟
🃠🃡🃢🃣🃤🃥🃦🃧🃨🃩🃪🃫🃬🃭🃮🃯
🃰🃱🃲🃳🃴🃵
```
image


be5invis:
Need to unify the thickness of the frame.

jmcwilliams403:
Better playing card back:
image

be5invis:
Later you could make the digits in the regular suites somehow taller

jmcwilliams403:
Progress:
image

be5invis:
You could make the letter and symbol slightly narrower for regular cards.

jmcwilliams403:
image

Issue #2427, opened 2024-07-22
Consider moving the digit `zero`‥`nine` CV feature tags in the next full/breaking version bump?


jmcwilliams403:
Digits `0`–`9` are currently almost at the back end of `cv01`–`cv99`. If any more CV features are added behind them, some or all of the digits may end up being pushed onto the more custom `VxAA`–`VxZZ` feature tags.

Perhaps it might be worth considering to move the range of CV features for digits `0`–`9` to a safe location in version 31 or thereabouts? Like either at the very beginning (at `cv01`–`cv10`) or alternatively right after the ASCII alphabetic characters (at `cv51`–`cv60`).

be5invis:
Wait till they really went to VX.
Resolve as by design.

be5invis:
OK I decide to do it now as it is a good time to do that.

PR #2428, opened 2024-07-23
Draft: Add variant selectors For Cyrillic Lower A/Ve.

jmcwilliams403:
#2427 was kind of foreshadowing what I wanted to do here.
```
!iIlL17|¦ ¢coO08BbDQ $5SZ2zs 96µm
CG6Qg9q¶ äöü Þẞðþſß ΓΔΛαδιλμξπτχ∂
ЖЗКУЭЯавжзклмнруфчъыьэя  <= != ==
```
example of tall cursive form:

ss03 upright:
image
ss03 italic:
image

example of disunified upright single storey vs. double storey for latin vs. cyrillic:

ss07 upright:
image
ss07 italic:
image

example of where in the original font the latin characters have italic forms but the cyrillic characters do not:

ss09 upright:
image
ss09 italic:
image

example of where it might be more tasteful to use a toothed or tailed design for the cyrillic letter when the latin letter uses a toothless design:

ss12 upright:
image
ss12 italic:
image

example of a convention to use a particular motion-serifed design under slab when the original font doesn't use a cursive ve but the font as a whole still uses italics otherwise:

ss18 slab upright:
image
ss18 slab italic:
image


be5invis:
SS03/italic is tall?

be5invis:
For now just add VSes using VX. I will reorganize them at v31.

jmcwilliams403:
> SS03/italic is tall?

yes, it's an example of tall cursive, based on Consolas:
image

> For now just add VSes using VX. I will reorganize them at v31.

That's what I did.

be5invis:
That's fine. It is good enough to merge now.

PR #2456, opened 2024-08-04
Add variant selectors for Greek lower Phi/Psi (Last for forseeable future).

jmcwilliams403:
In all reality, _this_ should be the very last time I file a PR for something like this; I can't even fathom adding anything else after this point, so I figured I'd file this PR to get it out of the way and then back off from adding anything else for a very long time.

Many fonts use the current varphi shape for normal phi and the other normal phi shape for varphi. The Unicode charts also formerly had their recommended glyphs swapped for a while very early on before later settling on the current configuration. LaTeX also has their respective names swapped to this day.

Additionally, there is also a "neo-hellenic" shape. Currently, it's only used by ss12, based on Ubuntu Mono, but other fonts that are not covered by Iosevka's ss features also use it, such as Discord's gg sans font for their UI.

For Psi's case, many fonts use the flat top design that this font uses for Cyrillic Psi. In fact, the Cyrillic character basically just forces that particular variant, but both are common in a Greek context.

All phi variants:
`φᵠᵩϕⱷ`
image

All psi variants:
`ψ𝞇ѱ`
image

```
G6Qg9q¶ Þẞðþſß ΓΔΛαβγδιλμνξπτυφχψ
```

Default for reference (unchanged):
image
ss01:
image
ss03 upright:
image
ss03 italic:
image
ss04:
image
ss05:
image
ss09:
image
ss12:
image
ss13:
image
ss14:
image
ss18:
image


dscorbett:
U+03D5 GREEK PHI SYMBOL is supposed to always have the full vertical stroke, regardless of the glyph of U+03C6. Consolas is wrong.

jmcwilliams403:
> U+03D5 GREEK PHI SYMBOL is supposed to always have the full vertical stroke, regardless of the glyph of U+03C6. Consolas is wrong.

Fixed.

Issue #2146, opened 2024-01-02
Curly `W`/`w` with a flat top.


jmcwilliams403:
- [x] The requested variant shape does not go too far away from Iosevka's design.
- [x] The requested variant does not conflict with any characters in Unicode that Iosevka currently supports.
- [x] At least two monospace/programming fonts, created by different designers, supported the requested variant. Provide images below.

------
`VvWv`

Mostly seen on low-resolution raster fonts:

FixedSys:
![image](https://github.com/be5invis/Iosevka/assets/37010132/1640995b-a125-4237-9925-4bdb81d314ab)
Press Start 2P:
![image](https://github.com/be5invis/Iosevka/assets/37010132/679ee5a5-5d4c-4314-a076-85a465b659f2)
VT323:
![image](https://github.com/be5invis/Iosevka/assets/37010132/5973f785-f556-461c-b781-4d5560630758)

But also able to be found on normal vector fonts:

DaddyTimeMono:
![image](https://github.com/be5invis/Iosevka/assets/37010132/00e13d94-5b76-4cf5-83c5-2982629af298)
ProFont:
![image](https://github.com/be5invis/Iosevka/assets/37010132/d96774d8-da9f-4723-9d02-afe290145839)

I'm also half-wondering if there's a case to be made for `W`/`w` with vertical sides and a flat top just to complete the set:

![image](https://github.com/be5invis/Iosevka/assets/37010132/3b25bcbb-ba8a-49de-8ce9-f52fbe042182)

be5invis:
It looks like you are requesting two things:

 * Curly `W`/`w` but with an extra center vertical bar reaching the top.
 * `W`/`w` with vertical sides and an extra center vertical bar reaching the top.

jmcwilliams403:
As mentioned in #2467 there also exists a rounded `w` with a flat top, looking like a "double u":
```
UVWY
uvwy
```
Nova Mono:
![image](https://github.com/user-attachments/assets/dddc28c6-f60a-4113-8d60-9fefaa24c16b)

PR #2467, opened 2024-08-13
Add `rounded-vertical-sides` variants for Capital/Lower W, for `ss10`.

jmcwilliams403:
This could also be added to #2146 because flat-top variants for this shape also exist.
However, as far as the current stylistic sets go, Envy Code R only uses this style.

After this I'm probably done adding variants for the time being. Anything else I can think of is either frivolous or outside of my current abilities.

```
G6Qg9q¶ Þẞðþſß ΓΔΛαβγδιλμνξπτυφχψ
```

ss10 sans upright:
image
ss10 sans italic:
image
ss10 slab upright:
image
ss10 slab italic:
image
all `W` variants:
`W𝖶Ⱳ`
image
all `w` variants:
`w𝗐ⱳ`
image


be5invis:
How many other variants are there to be added for SSes? I hope this is the last.

PR #2484, opened 2024-08-27
Add seven `ccmp` overrides, two `APLF` overrides, and two punctuation characters.

jmcwilliams403:
Six of the seven `ccmp` overrides are for glotallic vowels which were how these characters were represented before their formal unicode encoding. Early versions of their unicode proposal involved simply adding an annotation for cyrillic psili for its use for these characters. I do recall explicitly seeing fonts that do this override but I don't remember names.

The remaining override is for retroflex `q` which was apparently decided at some point that if a retroflex `q` were to ever be considered, it would be unified with `ɋ`. The capital version (`Ɋ`) needs no such override as it is officially an enlarged version of the lowercase, not a `Q` with a retroflex hook.

The two `APLF` overrides (not shown) are for UIUA for [above](https://www.uiua.org/docs/above) and [below](https://www.uiua.org/docs/below).

Zero Width Non-Breaking Space (not shown) is currently recommended for use as a byte order mark, and thus may either always appear at the beginning of a text file to show its byte order, or alternatively in its deprecated original purpose which is to mark a non-breaking point in the text stream (this use replaced by word joiner (`U+2060`) but it's recommended that the text stream handles the deprecated function if it appears anywhere else in the text but the beginning).

Kayah Li Sign Cwi is a punctuation character originally intended for the Kayah Li script but as of [L2/13-137](https://www.unicode.org/L2/L2013/13137-a92e-props.pdf) is now also used in Latin script, and is currently supported by Noto Sans/Serif/Mono (Latin) fonts as well as SIL's Andika/Charis/Doulos/Gentium/etc. fonts. Its appearance is similar to that of an inverted undertie but at half x-height.

```
A҆a҆I҆i҆U҆u҆ q̢
ꞺꞻꞼꞽꞾꞿ ɋ
 Vǎ puố̤vyá̤ thǐ꤮, bí thǐ myáhtye
sítôbè ná tè̤pyá̤tè̤sè̌ tè̤ilo-ilyá pwǒ̤꤮
tôcô hyǎbè thǐ akhè̌nuô, be è phú tè̤-
thè́krṳ̂̌ adu tôcô dố thǐgně nuô ní꤮.
```
image


packages/font-glyphs/src/space/index.ptl
```diff
@@ -71,3 +72,7 @@ glyph-block Spaces : begin
 		create-glyph [MangleName 'emsp'] [MangleUnicode 0x2003] : glyph-proc
 			set-width : if (para.isQuasiProportional && MosaicWidthScalar > 1) UPM MosaicWidth
 		alias [MangleName 'emquad'] [MangleUnicode 0x2001] [MangleName 'emsp']
+
```
be5invis, line 75: Please remove. It should follow the metrics of ideographs which Iosevka doesn't have.

packages/font/src/derive-spacing.mjs
```diff
@@ -85,6 +85,7 @@ async function deriveFixed_DropWideChars(font) {
 		0x27fe, // LONG RIGHTWARDS DOUBLE ARROW FROM BAR
 		0x27ff, // LONG RIGHTWARDS SQUIGGLE ARROW
 		0x2b33, // LONG LEFTWARDS SQUIGGLE ARROW
+		0x3000, // IDEOGRAPHIC SPACE
```
be5invis, line 88: Please remove.

be5invis (changes requested):

packages/font-glyphs/src/space/index.ptl
jmcwilliams403: It would need to be added anyway if #759 is added. It is the missing space character in that block.

jmcwilliams403:
Removing Ideographic Space for now but I would like to acknowledge that it would need to be added back in the future when approaching #759

be5invis (approved):

PR #2486, opened 2024-09-01
Add localized forms for Polish and Romanian.

jmcwilliams403:
Referenced from [this site](https://www.twardoch.com/download/polishhowto/intro.html).

Apparently Polish typography demands [transplanting the shape of the dot accent from e.g. lowercase `i`](https://www.twardoch.com/download/polishhowto/kropka.html) regardless of whatever `i`-specific optimizations it has on it.

Other languages that use `ż`, like Maltese and Irish, seem to tolerate (or even prefer, to some extent) leaving it distinguishable from the tittles of `i`/`j` (rather, only if it already is; i.e. not bothering to change it if it's not), since they use way more characters with dots above than just `ż` whereas `i`/`j` in particular seem to be outliers.

Kropka:
zdotaccent_iz

The acute accent should be more vertical, like a Greek Tonos but without the Greek-specific anchor point:
Kreska:
kreska_souvenir

Polish sample:
```
Pchnąć w tę łódź jeża lub ośm skrzyń fig.
PCHNĄĆ W TĘ ŁÓDŹ JEŻA LUB OŚM SKRZYŃ FIG.
```
Unlocalized for comparison:
image
Localized under `lang='pl'`
image

For Romanian, the hard-coded SComma and TComma characters were added to ISO-8859 and Unicode considerably late and thus a lot of existing text published up to that point had already been using the cedilla characters. As a result, [many fonts](https://localfonts.eu/typography-basics/fonts-the-importance-of-localisation/local-features/latin-local-features/) use glyph substitutions for `Ş`/`ş`/`Ţ`/`ţ` with `Ș`/`ș`/`Ț`/`ț` pointed at the hard Unicode indexes.

Romanian sample:
```
Înjurând pițigăiat, zoofobul comandă vexat whisky și tequila.
Înjurând piţigăiat, zoofobul comandă vexat whisky şi tequila. 
```
Unlocalized for comparison:
image
Localized under `lang='ro'`:
image


changes/31.5.1.md
```diff
@@ -0,0 +1 @@
+* Add Latin localization forms for Polish and Romanian.
```
be5invis, line 1: Change the version to 31.6.0. I want to better enforce SemVer -- given this is a new feature we should produce a minor version change.

be5invis (changes requested):

packages/font-glyphs/src/marks/above.ptl
```diff
@@ -1244,6 +1245,11 @@ glyph-block Mark-Above : begin
 			flat (markMiddle + d + wide) aboveMarkTop [heading Downward]
 			curl (markMiddle + d - wide) aboveMarkBot [widths.heading fine fine Downward]
 
+	create-glyph 'kreskaAbove' : glyph-proc
+		set-width 0
+		include : refer-glyph 'tonosAbove'
```
be5invis, line 1250: We should use a slightly more slanted shape for Kreska. Maybe for Tonos too.

be5invis (changes requested):

PR #2497, opened 2024-09-10
Random collection of geometric glyphs

Logo121:
None of these have real justification, just being close enough to existing glyphs.

`𝍠𝍡𝍢𝍣𝍤𝍥𝍦𝍧𝍨𝍩𝍪𝍫𝍬𝍭𝍮𝍯𝍰𝍱𝍲𝍳𝍴𝍵𝍶𝍷𝍸`
Pasted image 20240910225626
In QP:
Pasted image 20240910225643

`⎄ ⭘⭙ ⛚⛋⛝⛞⹙⹚⹛⹜`
Pasted image 20240910225713
</details>

---

Covered Range:
- `U+1D360` - `U+1D376` (The rest of counting rod numerals block)
	- These are straightforward to make. Mostly extension of the existing tally mark code.
	- For the Counting Rod numerals, the strokes were scaled to be the same across all the ones and tens digits.
		- this made the Heavy monospace strokes too thin, so the horizontal strokes are modified to use a different sw.
- `U+2E59` - `U+2E5C` (Half parenthesis)
	- Copied code for half square brackets.
- `U+2384`, `U+26DA`, `U+26CB`, `U+26DD`, `U+26DE`
	- Random collection of geometric symbols that can be done with some simple composition.
- `U+2B59` (Heavy Circled Saltire)
	- Implemented as "poweroff" with saltire for consistency. Ideally both should be actual circles, but Heavy Circle is also used as poweroff symbol, which Iosevka currently implements to be a standard "ring" instead.
	- If other symbols from the same set like `⭗` (or even `⭖`) are to be implemented, they may need to follow this as well, unless something is done with the power symbols.

### DIFF
--- a/packages/font-glyphs/src/symbol/counting-rod.ptl
+++ b/packages/font-glyphs/src/symbol/counting-rod.ptl
@@ -87,28 +87,6 @@ glyph-block Symbol-Counting-Rod : begin
 		create-glyph 'countingRodTens9' 0x1D371 : glyph-proc
 			include : rodTens 4 1
 
-	do "Tally Mark Ideographic"
-		create-glyph 'tallyIdeograph1' 0x1D372 : glyph-proc
-			local df : include : DivFrame rodDiv 3 1.1
-			include : df.markSet.capital
-			include : HBar.t df.leftSB df.rightSB CAP df.mvs
-		create-glyph 'tallyIdeograph2' 0x1D373 : glyph-proc
-			local df : include : DivFrame rodDiv 3
-			include : refer-glyph 'tallyIdeograph1' AS_BASE ALSO_METRICS
-			include : VBar.m df.middle 0 CAP df.mvs
-		create-glyph 'tallyIdeograph3' 0x1D374 : glyph-proc
-			local df : include : DivFrame rodDiv 3 1.4
-			include : refer-glyph 'tallyIdeograph2' AS_BASE ALSO_METRICS
-			include : HBar.m df.middle df.rightSB (CAP / 2) df.mvs
-		create-glyph 'tallyIdeograph4' 0x1D375 : glyph-proc
-			local df : include : DivFrame rodDiv 3
-			include : refer-glyph 'tallyIdeograph3' AS_BASE ALSO_METRICS
-			include : VBar.m [mix df.leftSB (df.middle - [HSwToV : df.mvs / 2]) 0.4] 0 (CAP * 0.55) df.mvs
-		create-glyph 'tallyIdeograph5' 0x1D376 : glyph-proc
-			local df : include : DivFrame rodDiv 3
-			include : refer-glyph 'tallyIdeograph4' AS_BASE ALSO_METRICS
-			include : HBar.b df.leftSB df.rightSB 0 df.mvs
-
 	do "Tally Marks"
 		create-glyph 'tallyMarkOne' 0x1D377 : glyph-proc
 			local df : include : DivFrame : if isProportional (rodDiv / 4) 1


jmcwilliams403:
Composition symbol (`⎄`) should always have the circle centered on the middle of the stroke of the edge of the square

packages/font-glyphs/src/symbol/counting-rod.ptl
```diff
@@ -0,0 +1,134 @@
...
+		create-glyph 'countingRodTens6' 0x1D36E : glyph-proc
+			include : rodTens 1 1
+		create-glyph 'countingRodTens7' 0x1D36F : glyph-proc
+			include : rodTens 2 1
+		create-glyph 'countingRodTens8' 0x1D370 : glyph-proc
+			include : rodTens 3 1
+		create-glyph 'countingRodTens9' 0x1D371 : glyph-proc
+			include : rodTens 4 1
+
+	do "Tally Mark Ideographic"
```
be5invis, line 92: Please remove. Iosevka is **not** a CJK typeface.

be5invis (changes requested):

changes/31.6.2.md
```diff
@@ -1 +1,11 @@
-- Fix Macedonian Cyrillic Gje under italics (#2493).
+* Fix Macedonian Cyrillic Gje under italics (#2493).
+* Add Characters:
```
be5invis, line 2: Please consider bump the version of the change file to 31.7.0 as there's new features added.

be5invis (changes requested):

Logo121:
I'll deal with the changes tomorrow, marking as draft now

> Ideographic tally marks should be forced to a wide monospace (technically duospace?) square grid.

Wasn't aware of this requirement. It _is_ marked as "Wide" in Unicode, but there are many other "Wide" glyphs that are implemented as Mosaic (both widths), or even just narrow width.

At least JuliaMono implements the whole Counting Rod block as halfwidth too, but again it's JuliaMono.

> Please remove. Iosevka is **not** a CJK typeface.

Again I'm not so sure where's the line between "CJK" and "non-CJK" for symbols like these. For me it's just another tally system, which some non-CJK but pan-Unicode font supports because they are straightforward enough to make.

But just to be clear, do I remove the Ideograph Tallies only, or the other counting rods as well? Since their primary usage is in CJKV anyway.

be5invis:
The ideographic tally marks are partial writing of the "正" (U+6B63) character.
See https://www.unicode.org/L2/L2016/16046-ideo-tally-marks.pdf.

> But just to be clear, do I remove the Ideograph Tallies only, or the other counting rods as well? Since their primary usage is in CJKV anyway.

Remove the five ideographic tally marks only.

Logo121:
Ok the current revision should have the widths fixed now (the unspaced Tens glyphs looks like that on my computer, probably Noto. Nishikiteki also has this form iirc). The other changes (including the deletion) are applied accordingly.

But for monospaced version, I'm not sure what you mean by "Default them to Wide". Since they are not mosaics at the moment (following the Tally Marks), only Narrow version exists.  
So do I:
- Make the Counting Rod mosaics (but keep the behavior of Tally Marks as is);
- Make the whole block mosaic, but somehow keep the Tally Mark One width for QP only (e.g. making them not Mosaics in QP); or
- Make the Counting Rod Wide only (like the long arrows) and have them removed in Monospace-only versions?

---

> The ideographic tally marks are partial writing of the "正" (U+6B63) character.
> See https://www.unicode.org/L2/L2016/16046-ideo-tally-marks.pdf.

This part I know, but my point is when other font creators make glyphs for this block (which admittedly there are probably less than 10 fonts supporting the block completely), I doubt most would even care about those being CJK or not because they are easily abstractable as just "5 orthogonal lines in a certain pattern".

I'm also not quite aware of any technicalities or mechanisms that would make this font "CJK font" just because of the inclusion of these glyphs, unlike the squared units last time (#1651) or the ideographic space. Unless the "Ideographic" in its name is somehow triggering something?

jmcwilliams403:
> I'm also not quite aware of any technicalities or mechanisms that would make this font "CJK font" just because of the inclusion of these glyphs, unlike the squared units last time (ref) or the ideographic space. Unless the "Ideographic" in its name is somehow triggering something?

Belleve has said in the past that inclusion of certain characters whose primary script is officially specified as something ideographic causes certain applications to treat the whole font as a CJK font, and if that were the user's actual intention then it would be better to use [Sarasa Gothic](https://github.com/be5invis/Sarasa-Gothic) instead, which includes Iosevka as a subset.

be5invis:
> Default them to Wide

nvm I will implement this. LGTM now.

PR #2499, opened 2024-09-12
Add more character overrides to two Latin locales.

jmcwilliams403:
Firstly, removal of serif variants for specifically the _Latin_ version of Capital/small Schwa unless under a supported Turkic locale, namely Azerbaijani whose usage of schwa comes directly from formerly using the Cyrillic script. This also incidentally returns it to its behavior from before engineering serif variants onto it (the serifs of the Cyrillic version are unchanged)

Secondly, removal of crossbar variants for Latin `Z`/`z` when under a Polish locale because, much like the reason Greek Zeta doesn't use crossbars, it makes it confusable with the handwritten form of another letter, which in this case is `Ż`/`ż` (whereas for Greek it's `Ξ`). Letters not used in Polish such as `Ž`/`ž` etc. are not affected.

Thirdly, make {super|sub|over|under}script forms of schwa, `z`, and `i` respond to localization overrides.

### DIFF
--- a/changes/31.7.0.md
+++ b/changes/31.7.0.md
@@ -1,7 +1,7 @@
 * Fix Macedonian Cyrillic Gje under italics (#2493).
 * Improve widths of overline marks of Serbian italic lower Ghe/Pe/Te.
 * Remove crossbar variants for `Z`/`z` when under Polish locale (`PLK`) to avoid confusion with the handwritten `Ż`/`ż` allograph, `Ƶ`/`ƶ`.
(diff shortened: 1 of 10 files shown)


changes/31.7.0.md
```diff
@@ -1,5 +1,7 @@
 * Fix Macedonian Cyrillic Gje under italics (#2493).
 * Improve widths of overline marks of Serbian italic lower Ghe/Pe/Te.
+* Remove crossbar variants for `Z`/`z` when under Polish locale (`PLK`).
```
be5invis, line 3: Comment why doing this.

changes/31.7.0.md
```diff
@@ -1,5 +1,7 @@
...
+* Make serif variants for Latin Capital/Small Schwa (`Ə`/`ə`) only appear under Turkic (Turkish/Azerbaijani/etc.) locales (`TRK`, `AZE`, `GAG`, `KAZ`, `TAT`, `CRT`).
```
be5invis, line 4: Again comment why removing them.

be5invis (changes requested):

be5invis (approved):

PR #2540, opened 2024-10-08
Random collection of more geometry characters

Logo121:

```
⌯⌲⌳⌇⸾⸽
⦚⧘⧙⧚⧛⸽
⟠⟘⟙⨅⨆⸽
```
image

</details>

---

- `U+232F` (`⌯`)
	- Dimension specification symbol (like `⌀`). Implemeted as 3 Horizontal Strokes.
- `U+2332` - `U+2333` (`⌲⌳`)
	- Dimension specification symbol. Implemented using `TriangleShape`
	- Following the shape [here](https://www.gdandtbasics.com/gdt-symbols/) which differs a bit from Unicode reference (and Noto). Not sure if this is needed, but at least Unifont seems to do this too, fwiw.
- `U+27E0` (`⟠`)
	- Another masked composition. Apparently used for representing Ethereum.
- `U+2307` (`⌇`)
	- Adopted "Sine Wave" implementation, but vertical
- `U+2E3E` (`⸾`)
	- Further modified from Wavy Line. The glyph still gets a bit deformed in extreme cases (max slant max width max weight), not sure how to make it better (other than further thinning the stroke width).
	- Same height as `|`, since they are related.
- `U+2E3D` (`⸽`)
	- 6 dots. Same height as `|` again.
- `U+29D8` - `U+29DB` (Zigzag lines `⦚⧘⧙⧚⧛`)
	- Generalization of the vertical resistor, which is a shared shape now.
	- Same height as `()` as most of them are bracket pairs.
- `U+27D8`, `U+27D9`, `U+2A05`, `U+2A06` (Large math ops `⟘⟙⨅⨆`)
	- Extension of existing glyphs 
	- Due to how the existing glyphs are implemented bigsqcup and bigsqcap will look identical to product and coproduct in Sans-serif. Leaving it as is until there are better solutions.

Other file changes are minor wording fixes, replacements with existing variables/functions, etc.


jmcwilliams403:
> Due to how the existing glyphs are implemented bigsqcup and bigsqcap will look identical to product and coproduct in Sans-serif. Leaving it as is until there are better solutions.

FWIW this is only a coincidence/technicality. Product/Coproduct have serifs because they are based on a literal large Greek Pi whereas square cups/caps simply resemble it by accident. There are also plenty of other math characters which are homoglyphs of others, either sometimes or always.

I think it's fine the way you have it.

changes/31.8.1.md
```diff
@@ -6,3 +6,17 @@
 * Optimize glyph for Cyrillic Lower Dzze (`U+A689`) under italics.
 * Optimize glyphs for Volapük Ae/Oe/Ue (`U+A79A`..`U+A79F`).
 * Optimize glyph for Latin Lower Dezh Digraph with Palatal Hook (`U+1DF12`).
+* Add characters:
```
be5invis, line 9: Since now we added new featurs, please bump the version to 31.9.0 as we are better enforcing semver. Rename this file to `31.9.0.md`.

be5invis (changes requested):

be5invis:
And please optimize the stroke width of `⌲` and `⌳` in narrow width & heavy weight.

Logo121:
Ultracondensed:
image

Using AdviceStroke2. I feel like if i increase the crowdedness a bit more the lighter weights will start look oddly thin.

be5invis (approved):

PR #2552, opened 2024-10-15
Draft: Armenian letters

### LINKED ISSUE #1664: Block Request: Armenian (`U+0530..U+058F`)
  - [ ] Latin
  - [ ] Cyrillic
  - [ ] Greek
  - [ ] Punctuation
  - [ ] Symbol

------

The Armenian script is like a long-lost relative of Latin/Greek/Cyrillic, and is often also displayed under a Roman font, as glyphs can be indirectly derived from lookalikes in Latin/Greek/Cyrillic

Example in Consolas:
image
Example in Courier New:
image
Example in Julia Mono:
image

Logo121:
<details> <summary> images </summary>

imageimage

</details>

Preliminary effort to resolve #1664.

Note:
- All of them are put into the same file for now (unless there are external modifications), but can be split once the implementations are ready.
	- Some "common components" are implemented as their own function, mostly for semantic(?) purposes.
- Based on Arian AMU, the [image reference with lines in Wikipedia](https://commons.wikimedia.org/wiki/Category:Armenian_upper-_and_lower-case_pairs_with_font_lines), and some other Armenian fonts on fonter.am if the former 2 is not enough.
- Some dimensions are to be tuned.
- This implementation does not consider (or at least put too much effort in):
	- Character variants (far far future concern if even planned).
		- For characters that do have variants, usually the simplest, most geometric-like variant is picked.
		- Since some glyphs are direct copies/references of other LGC glyphs, they may be affected by variants.
		- Only consider the SLAB/sans-serif difference for now.
	- Italic forms
		- There might be some in lowercase, or even on uppercase if we use the handwriting forms for italic. Still needs investigation.
		- Some serif differences are taken care of, but not comprehensive (e.g. serifs turning into tails).
	- Other specific metrics like QP width, leaning marks etc.
- Only letters are implemented now, so these are not supported yet (in order of difficulty):
	- Punctuation (including the Currency symbol)
	- Diacritics
	- The Eternity symbol
- Known problems:
	- (lower xe `խ` is wrong in the visualization. this has been fixed in the code but the screenshot is before the fix.)
	- The "juts" in some of the letters extrudes outside the character cell and overlaps with adjacent characters.
	- On the other hand, sometimes they may look too similar to serifs, if any.
	- The curves in some letters may look awkward, due to the complexity of the curve itself and/or the narrow metric of Iosevka. 
	- Serif overlaps in some weights/width configurations.
- No changelog yet. But it's basically `U+0531 ... U+0556` and `U+0560 ... U+0588`.

Other file changes:
- `upper-q.ptl`:
	- Exported the straight crossing bar for Ayb.
- `shared.ptl`:
	- Expanded SerifFrame to support more `mt` stuff.
	- All OBar shapes are given a `bot` parameter. The OBarRight ones are not tested yet.
- `shapes.ptl`:
	- Added a new `flatside` function, basically replicating and generalizing the "arc depths larger than height" cases in OShape. This is applied to OBar shapes.
-  `macros.ptl`:
	- Adding `flatside` to the default imports.

### DIFF
--- a/packages/font-glyphs/src/letter/latin/lower-a.ptl
+++ b/packages/font-glyphs/src/letter/latin/lower-a.ptl
@@ -275,10 +275,11 @@ glyph-block Letter-Latin-Lower-A : begin
 	derive-composites 'aRetroflexHook' 0x1D8F 'a/rtailBase'
 		RetroflexHook.rSideJut RightSB 0 (yOverflow -- Stroke)
 
-	select-variant 'scripta' 0x0251
-	select-variant 'largescripta' 0x2C6D (follow -- 'scripta')
-	select-variant 'invscripta' 0xAB64 (follow -- 'scripta')
-	derive-composites 'alphaRetroflexHook' 0x1D90 'scripta.singleStoreySerifless'
+	select-variant 'scripta'      0x251  (follow -- [conditional-follow SLAB 'scripta/autoSerifed/slab' 'scripta/autoSerifed/sans'])
+	select-variant 'largescripta' 0x2C6D (follow -- [conditional-follow SLAB 'scripta/autoSerifed/slab' 'scripta/autoSerifed/sans'])
+	select-variant 'invscripta'   0xAB64 (follow -- [conditional-follow SLAB 'scripta/autoSerifed/slab' 'scripta/autoSerifed/sans'])
+
+	derive-composites 'scriptaRetroflexHook' 0x1D90 'scripta.singleStoreySerifless'
 		RetroflexHook.rSideJut RightSB 0 (yOverflow -- Stroke)
 
 	CreateTurnedLetter 'turnscripta' 0x252 'scripta' HalfAdvance (XH / 2)
--- a/packages/font-glyphs/src/letter/latin/lower-g.ptl
+++ b/packages/font-glyphs/src/letter/latin/lower-g.ptl
@@ -207,7 +207,6 @@ glyph-block Letter-Latin-Lower-G : begin
 	link-reduced-variant 'g/sansSerif' 'g' MathSansSerif
 	CreateTurnedLetter 'turng' 0x1D77 'g' HalfAdvance [mix Descender XH 0.5]
 	select-variant "gBar" 0x1E5 (follow -- 'g')
-	select-variant 'g/single' null (shapeFrom -- 'g') (follow -- [conditional-follow SLAB 'g/single/autoSerifed/slab' 'g/single/autoSerifed/sans'])
 
 	select-variant 'g/hookTopBase' null (shapeFrom -- 'g')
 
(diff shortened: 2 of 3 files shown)


be5invis:
Some suggestions:

1. Separate the Armenian file into separate files, group by letters' shape.
2. Prefix glyph names unique for Armenian with `armn/` or `Armn/`.

Letter-specific suggestions:

* `Փ`: should fit in cap height.
* `Ց` `Ֆ` `ֆ`: make the crossing placed closer to the center.
* `ք`: consider raise the circle's  bottom edge and make the horizontal bar place at baseline. The horizontal bar should not be that far at right side.
* `Ջ`: We need better detachment of the ring part and the bottom part. Consider take the design from Noto.
* `ե`, `է`: the middle horizontal bar could stop at `RightSB` for its right end.
* `Ճ`: Could be less "square-y".
* `Հ`: Could be made closer to a L shape so it could be easier distinguished with `Յ`.
* `յ`, `ֈ`: The hook should be less significant -- consider use the flat-hook shape we had in `dotlessj`.

Logo121:
> 1. Separate the Armenian file into separate files, group by letters' shape.

Will do probably after everything is fixed.

>  2. Prefix glyph names unique for Armenian with `armn/` or `Armn/`.

I think I did already? What did I miss?

> * `Փ`: should fit in cap height.

I just realized `cyrl/Ef.BGR` is not in cap height. Changing to alias of `grek/Phi`.

> * `Ց` `Ֆ` `ֆ`: make the crossing placed closer to the center.

Realigning to HBarPos (Same as Ho).

> * `ք`: consider raise the circle's  bottom edge and make the horizontal bar place at baseline. The horizontal bar should not be that far at right side.

Done (following Arian AMU)

> * `ե`, `է`: the middle horizontal bar could stop at `RightSB` for its right end.

Done (following Arian AMU). In return the bottom horizontal bar for `է` is shortened, because that's how it looks like in most fonts

> * `յ`, `ֈ`: The hook should be less significant -- consider use the flat-hook shape we had in `dotlessj`.

Ok. Not sure if my change is the correct way to handle "different aliases depending on SLAB setting" though.

> * `Հ`: Could be made closer to a L shape so it could be easier distinguished with `Յ`.

image
I took this to mean that the R-like form is the more LGC-like shape, or at least the most fitting to Iosevka's style. This is the form used in Arian AMU, which *is* a neo-grotesque font (based on Arial).

> * `Ճ`: Could be less "square-y".

image
This is the form in Arian AMU.

> * `Ջ`: We need better detachment of the ring part and the bottom part. Consider take the design from Noto.

I kinda got stuck on this (this character took me the longest to implement in the first place). The design itself isn't very different from Noto Armenian (ignoring the downward juts), it's just the metrics and probably my spiro-fu not being good enough to make this not break randomly in some settings.
image
This is how it looks for now.

be5invis:
I think for `ք` the bottom bar shall be fit in LeftSB-RightSB space.

be5invis:
For `Ֆ` `ֆ` maybe we could make the top-left contour horizonally larger (like having the same horizontal size as the bottom-right contour).
image

be5invis:
For `Ջ` you can maybe make the vertical serif at bottom-left thinner.

be5invis:
For `Թ` we could probably make the outer conour still uase `Stroke` as stroke width. Only make the inner curl thinner.

Logo121:
whoops, made a mistake. reopening a different pr...

PR #2556, opened 2024-10-19
Draft: Armenian Letters (again)

Logo121:
Reopening of #2552. Accidentally discarded when trying to merge the new commits to the branch.

<details> <summary> images </summary>

image image

</details>

----

Things changed since last push:
- Split the characters roughly into groups by their shape.
  - The remnants of the original file is in `other.ptl`, though that only contains aliases now. To be cleaned up.
- Improved the italic serifs for n/u-shaped letters a little bit: now they follow the motion-serif-like pattern of topleft + bottomright outer serifs.
- Added the currency symbol Dram and the full stop (aliased to colon).
- As hinted by #2553, using single-storey `g` (aliased to bulgarian de) for lower Co. 

For the suggested changes:
> I think for `Ք` `ք` the bottom bar shall be fit in LeftSB-RightSB space.
> You could probably place the top edge of `ք`'s bar at baseline. Balance the spaces.

I only moved the right part of the bar (forgot to do so for upper Keh) for now since the left part should extrude out of the stem. Should I also move left stem rightward?

For the second point, doesn't seem to be possible for ultraextended heavy slab, which the bar will overlap with the bottom serif. I moved it to the center at baseline for now.

> For `Ֆ ֆ` maybe we could make the top-left contour horizonally larger (like having the same horizontal size as the bottom-right contour).

Done (but added overshoot for the hook part so that it looks a little bit better).
image

> For Ջ you can maybe make the vertical serif at bottom-left thinner.

Also done (also for Za for consistency). 
image

> For `Թ` we could probably make the outer conour still uase `Stroke` as stroke width. Only make the inner curl thinner.

Barely. 
image
With normal stroke the space left in the middle is very small, so the inner curl would be really thin to fit in.

It still works in the ultracondensed case for now:
image


be5invis:
Can you try typeset some text for sample? Also when you take screenshot you could have higher resolution :)

Logo121:
Sample text (forgot where I got this):

```
Նափոլէոն կարդացած է Քորնէյի,
Ռասինի եւ Վոլթերի ստեղծագործութիւնները,
իսկ անոր սիրելի բանաստեղծը Օսսիանն էր։
```
<details> <summary> higher resolution </summary>

image
image

QP (less polished)
image

</details>

be5invis:
We may need some extra touching for `թ` which could make the inner stroke thinner, sort of like its uppercase.
Need some balancing for `լ` and `յ` at monospace -- Slightly shift them to left or right if necessary. You can follow the shift number of implementation of `l.flatTailed` or `j.flatHook{Serifless | Serifed}`.

be5invis:
Consolas have some Armenian that you could use as reference:

image
image

Logo121:
> We may need some extra touching for `թ` which could make the inner stroke thinner, sort of like its uppercase.

`թ` already has the same inner stroke width as the uppercase. Should I make it even thinner?

>  Need some balancing for `լ` and `յ` at monospace -- Slightly shift them to left or right if necessary. You can follow the shift number of implementation of `l.flatTailed` or `j.flatHook{Serifless | Serifed}`.

`յ` is already aliased to dotless `j`, so i don't think it should have problems.

Changed `լ` to use LongJut (the same way hookybottom `l` does).

Logo121:
<details> <summary> image </summary>

image
image

</details>

be5invis:
Nice, this looks a lot better.
The space between `րծ` looks a little large.

packages/font-glyphs/src/letter/armenian/other.ptl
```diff
@@ -0,0 +1,121 @@
...
+	alias 'armn/co' 0x581 'cyrl/de.BGR'
+
+	## From phi:
+	alias 'armn/Piur' 0x553 'grek/Phi'
+
+	## From o:
+	alias 'armn/Oh' 0x555 'O'
+	alias 'armn/oh' 0x585 'o'
+
+	do "Ayb" # Upper U Group
```
be5invis, line 49: I prefer remove move the empty "do"s here.
You can leave comments in `letter/armenian.ptl`, indicating which letter lives in which file. For example
```ptl
    * Ayb U+0531      => letter/armenian/upper-u-group.ptl
    * Ben U+0532       => letter/armenian/hook-group.ptl
```

packages/font-glyphs/src/letter/armenian/ca.ptl
```diff
@@ -0,0 +1,67 @@
...
+
+			include : intersection [MaskBelowLine (x1 - xOffset) Ascender (x2 - xOffset) y2 100] : dispiro
+				widths.rhs df.mvs
+				flat (x3 + xOffset) Ascender
+				curl (x4 + xOffset) XH
+				# arcvh
+				straight.down.end (df.rightSB + OX) (XH - SmallArchDepthB)
+				curl (df.rightSB + OX) (0 + SmallArchDepthA)
+				arch.rhs 0 (sw -- df.mvs)
+				flat (df.leftSB - OX) (0 + SmallArchDepthB)
```
be5invis, line 62: OX is negative, so, you need to *add* OX for left edge and *minus* OX for right edge.

be5invis (changes requested):

packages/font-glyphs/src/letter/armenian/ca.ptl
be5invis: I may need to refactor the code to make `O` and `OX` positive, so it could be more intuitive for contributors.

packages/font-glyphs/src/letter/armenian/other.ptl
Logo121: This whole file is a leftover to be cleaned up anyway, I'll just move all the actual aliases to their corresponding sources and remove this file.

Logo121:
image

packages/font-glyphs/src/letter/latin/lower-g.ptl
```diff
@@ -217,6 +217,7 @@ glyph-block Letter-Latin-Lower-G : begin
 
 	select-variant 'cyrl/de.BGR' (shapeFrom -- 'g') (follow -- [conditional-follow SLAB 'g/single/autoSerifed/slab' 'g/single/autoSerifed/sans'])
 	alias 'cyrl/de.SRB' null 'cyrl/de.BGR'
+	alias 'armn/co' 0x581 'cyrl/de.BGR'
```
be5invis, line 220: I suggest putting this into `letter/armenian` but create a new file (say, `letter/armenian/aliases.ptl`)

be5invis:
> image

This looks much better :)

packages/font-glyphs/src/common/shapes.ptl
```diff
@@ -530,6 +494,36 @@ glyph-block CommonShapes : begin
 		local args : object [yRef y] sw swTerminal isTail noSwash [overshoot o]
 		return : WithKnotProxy [hookProxy args] : new FunctionInterpolator hookEndBlender args
 
+	glyph-block-export flatside
+	define flatside : namespace
+		export : define [ld l d u ada adb _af] : return
```
be5invis, line 499: If we are adding `flatside` to build the sides for circle-likes, then maybe we could perform overshoot processing inside these functions, as the function names already contain the information about which side the stroke is. An exmaple could be:
```
...
export define flex-params [ld]: begin
    local-parameter : x # X-position of the stroke, without overshoot
    local-parameter : d # Y-position of the ring's bottom, without overshoot
    local-parameter : u # Y-position of the ring's top, without overshoot
    local-parameter : af -- nothing # Additional function for the control knots
    local-parameter : ada -- ArchDepthA
    local-parameter : adb -- ArchDepthB
    local-parameter : overshoot -- OX


    if (u - d > ada + adb)
        list
            flat (x + overshoot) (u - ada) af
            curl (x + overshoot) (d + adb)
        g4 (x + overshoot) [mix d u (adb / (ada + adb))] af
```

Logo121:
hopefully this doesn't break anything...

be5invis:
> hopefully this doesn't break anything...

I think they are fine.

Logo121:
Edited changelog, though I suppose the version would probably be bumped.

I suppose the main issues are fixed and it's ready to be merged, if there are no other problems.

changes/31.9.2.md
```diff
@@ -1,4 +1,7 @@
 * Add characters:
+  - ARMENIAN CAPITAL LETTER AYB (`U+0531`) ... ARMENIAN CAPITAL LETTER AYB (`U+0556`).
```
be5invis, line 2: It worth a seprate item :)

```markdown
* Added Armenian letters:
  - ARMENIAN CAPITAL LETTER AYB (`U+0531`) ... ARMENIAN CAPITAL LETTER AYB (`U+0556`).
  - ...
```

be5invis (changes requested):

be5invis:
> Edited changelog, though I suppose the version would probably be bumped.
> 
> It's probably ready to be merged, if there are no other problems.

Right, bump the minor version.

packages/font-glyphs/src/letter/armenian/lower-q-group.ptl
jmcwilliams403, line 1: Using `include : refer-glyph 'q'` when the original `q` character is subject to CV features is not good practice. It should instead select its appropriate variants from within `latin/lower-q.ptl`. Otherwise it will appear as if it were locked to whatever `q` variant the user compiles their font with by default, which also might not be one you expect here.

packages/font-glyphs/src/letter/armenian/lower-u-group.ptl
```diff
@@ -0,0 +1,347 @@
...
+				include sf.lt.outer
+				if [not para.isItalic] : begin
+					local sf2 : SerifFrame.fromDf df XH (XH / 2)
+					include : composite-proc sf.lb.full sf2.rb.full
+
+	do "Da"
+		create-glyph 'armn/da' 0x564 : glyph-proc
+			local df : include : DivFrame 1
+			include : df.markSet.p
+			include : refer-glyph 'latn/eta'
```
jmcwilliams403, line 44: This should be moved to `latin/lower-n.ptl` to have more consistent variant behavior. BTW `ss17` uses a tailed `n` under italics which might not be what you want here.

Logo121:
Well... yeah, I'm kinda aware of the compatibility problem with variants (see top comment in #2552).

I reckon it will be hard to reliable "link" Armenian letters to some existing character variants, because most characters that share a shape with Latin letters are not actually related to them etymologically(?), and even if we just link by shape, there will be a few characters where the exact character it should link to is ambiguous.

I'd say just merge it as is first (unless there are actual problems where the variant shape would affect the derived character -- I fixed it for `q`-based ones at least), then handle the variants linking/unlinking later. You can do it if you want.

packages/font-glyphs/src/letter/armenian/lower-u-group.ptl
Logo121: Well, if the wiki images for Armenian handwriting is any indication, it's probably fine to have the tail in italic (not that I'm trying to replicate that).

Lower ra is already differentiated by having the n shoulder shift inward.

be5invis (approved):
I think we are good enough.

PR #2557, opened 2024-10-20
Optimize serifed `semi-chancery` variants for `x`.

jmcwilliams403:
Inspired by code for `Ƙ` where it corrects for the ending x-coordinate of the hooked spur when the other leg is serifed.
`x𝗑хҳԕԗ`
Before:
image
After:
image

### DIFF
--- a/packages/font-glyphs/src/letter/latin/x.ptl
+++ b/packages/font-glyphs/src/letter/latin/x.ptl
@@ -63,31 +63,22 @@ glyph-block Letter-Latin-X : begin
 
 		local fine : [AdviceStroke 3] * (sw / Stroke)
 
-		local leftXExt  : leftX  - SideJut
-		local rightXExt : rightX + SideJut
+		local leftXExt  : leftX  - [if fSlab SideJut 0]
+		local rightXExt : rightX + [if fSlab SideJut 0]
 
 		return : dispiro
 			if fHalf
 				flat [mix leftX rightX 0.5] [mix leftY rightY 0.5] [widths.center fine]
-				if fSlab
-					list
-						flat leftXExt (leftY - sign * sw / 2) [widths.center.heading sw Rightward]
-						curl (leftX + TINY) (leftY - sign * sw / 2) [widths.center.heading sw Rightward]
-						alsoThruThem {{blendK1X blendK1Y} {blendK2X blendK2Y}}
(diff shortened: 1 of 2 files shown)


be5invis:
Look at cv58#13, why are the letters become wider?

jmcwilliams403:
> Look at cv58#13, why are the letters become wider?

Basically it makes a guess that the second chancery counter leg is covering up a serif, so under slab it adds a jut to both legs. Sans is unchanged, as shown by the mathematical sans serif after it.

packages/font-glyphs/src/letter/latin/x.ptl
```diff
@@ -196,7 +199,7 @@ glyph-block Letter-Latin-X : begin
 		cursive                         { STROKE-CURSIVE  STROKE-CURSIVE  null                     false }
 		semiChanceryStraightSerifless   { STROKE-STRAIGHT STROKE-CHANCERY null                     false }
 		semiChanceryCurlySerifless      { STROKE-CURLY    STROKE-CHANCERY null                     false }
-		chancery                        { STROKE-CHANCERY STROKE-CHANCERY null                     false }
+		chancery                        { STROKE-CHANCERY STROKE-CHANCERY null                     SLAB  }
```
jmcwilliams403, line 202: @be5invis would you like me to undo this line?

packages/font-glyphs/src/letter/latin/x.ptl
```diff
@@ -63,6 +63,9 @@ glyph-block Letter-Latin-X : begin
 
 		local fine : [AdviceStroke 3] * (sw / Stroke)
 
+		local leftX  : _leftX  - [if fSlab SideJut 0]
```
be5invis, line 66: I think we should only extend the chancey strokes to "cover" serifs only if there's serif. So for cv58#13 we should use the original width.

And maybe when we do extension, we should not modify the position of the curved turnings. We should perhaps create a new variant of `leftX`/`rightX` value that adds `SideJut`, and modify the flat ends' code into something like

```ptl
     ....
     flat (rightX - TINY) (rightY + sign * sw / 2) [widths.center sw]
     curl rightXExt (rightY + sign * sw / 2) [widths.center sw]
```

be5invis (changes requested):

jmcwilliams403:
image

PR #2604, opened 2024-11-30
Make six characters slightly wider under Quasi-Proportional.

jmcwilliams403:
The counting rods are hard-clamped to a diversity of 2/1 but they do not currently exceed that boundary; only meeting it exactly.
The rest are T-derived or T-convergent characters and Archaic M.
For bandwidth reasons, I'm only showing screenshots of upright forms as the italic forms of Cyrillic Te-derived characters are unchanged.
```
𝍠𝍡𝍢𝍣𝍤𝍥𝍦𝍧
𝍨𝍩𝍪𝍫𝍬𝍭𝍮𝍯
𝍰𝍱
𝍷𝍷𝍷𝍷𝍸
ТЂꚊЋꙐЪЫЬ
тђꚋћꙑъыь
```
Monospace for comparison (mostly unchanged except for that I fixed an issue with the stroke width of `Ꚋ`/`ꚋ` AKA TeMidHook under Heavy):

Sans Thin:
image
Sans Regular:
image
Sans Heavy:
image
Slab Thin:
image
Slab Regular:
image
Slab Heavy:
image

Quasi-Proportional before and after:

Aile Thin Before:
image
Aile Thin After:
image
Aile Regular Before:
image
Aile Regular After:
image
Aile Heavy Before:
image
Aile Heavy After:
image
Etoile Thin Before:
image
Etoile Thin After:
image
Etoile Regular Before:
image
Etoile Regular After:
image
Etoile Heavy Before:
image
Etoile Heavy After:
image

Issue #2182, opened 2024-01-26
Cyrillic letter `ы` is too wide


dom1n1k:
In **Aile** and **Etoile** variants cyrillic `ы` is unnaturally wide.
Optically, `ы` should be approximately equal to `ш` or `ю`.
But it's definitely wider and looks like two separate letters.

![github-iosevka](https://github.com/be5invis/Iosevka/assets/1683168/e5ca0368-13ac-4b3a-827c-5daeffbfe7b2)

---

* Your font version: 28.0.6
* Your font variant: Aile and Etoile
* Your operating system (name and version): Windows 10
* Your application using Iosevka: Figma

PR #2605, opened 2024-12-01
Make two characters slightly wider under Quasi-Proportional.

jmcwilliams403:
Continuation of #2604 .

Also fix changelog from #2604 as well; I forgot to remove the line mentioning the counting rods despite having returned them to normal as requested.

`ѦѨѧѩѪѬѫѭ`

Monospace for comparison (unchanged):

Sans Thin:
image
Sans Regular:
image
Sans Heavy:
image
Slab Thin:
image
Slab Regular:
image
Slab Heavy:
image

Quasi-Proportional before and after:

Aile Thin Before:
image
Aile Thin After:
image
Aile Regular Before:
image
Aile Regular After:
image
Aile Heavy Before:
image
Aile Heavy After:
image
Etoile Thin Before:
image
Etoile Thin After:
image
Etoile Regular Before:
image
Etoile Regular After:
image
Etoile Heavy Before:
image
Etoile Heavy After:
image


Logo121:
Oh, I've thought about the widths of the Iotated Yus and some other horizontally packed characters before.

It may be useful to add a width class beyond M (`M2`?) for characters that augment an already M-width character (or something equally crowded) e.g.:
- `ѨѩѬѭ`
- `ꚅ` and `ԫ` (the latter actually had slightly overlapping strokes in UltraCondensed at one point but was fixed with some width changes) 
- `ↈ` and `ꙮ` (also high amount of vertical strokes)
- `⋙⋘` (maybe)

I wasn't exactly sure this was necessary (especially since the overlapping problem is already fixed), but if we're doing this `[mix 1 diversityM 2]` thing for more characters now, it might be a good idea to just establish a new class like that `diversityT` a while back.

jmcwilliams403:
> Oh, I've thought about the widths of the Iotated Yus and some other horizontally packed characters before.
> 
> It may be useful to add a width class beyond M (`MM`?) for characters that augment an already M-width character (or something equally crowded) e.g.:
> I wasn't exactly sure this was necessary (especially since the overlapping problem is already fixed), but if we're doing this `[mix 1 diversityM 2]` thing for more characters now, it might be a good idea to just establish a new class like that `diversityT` a while back.

I was thinking about that but more for a diversity of `1.5` for compound characters including `W`/`w` and lowercase `m`. I did some experements recently and many characters look great but then characters like `Ы`/`ы`/`Ю`/`ю` would need major retooling, so I scrapped it.

Logo121:
I think diversityM is already good enough for `m` and the other characters you've mentioned. (in fact, there's already an issue complaining `ы` was too wide at one point: #2182)

The (potential) problem is with characters that compound with characters that are already diversityM, like Yus here.

jmcwilliams403:
> In fact, there's already an issue complaining `ы` was too wide at one point

I believe that was in comparison to `ш`/`m`. If they stay the same relative to each other then the issue remains solved.

jmcwilliams403:
I think I finally figured my approach out and the solution was to simply _not_ expand yeru or yu any more than they already are, and focus on sha/m/etc., making the former follow `diversityM` and the latter follow `diversityMM`.
image
image
image
image

jmcwilliams403:
@be5invis is this worth a PR? Maybe tomorrow?

be5invis:
@jmcwilliams403 You can do it

jmcwilliams403:
Okay I'll file a PR tomorrow. The code is basically ready but I just want to tidy some things up.

PR #2618, opened 2024-12-17
Adding/Harmonizing some symbols with small circles

Logo121:

`(♭𝄬𝄭♮𝄮𝄯♯𝄰𝄱)`

`⚬⚭⚮⚯`

`⯺⯻⯼⯽⊶⊷⧟⟜⊸∘`

</details>

----
- Changes in `arrows.ptl` (supposedly no glyphs are really affected):
	- Roughly regroup the top-level functions into "Arrow Bars", "Arrow Head/Tails", "Arrow Shapes", "Special Arrow Shapes" (for black/white and rounded arrows), "MkArrow Functions", denoted by comments for now
	- Added the optional argument `_halfSw` to all Arrow Shape functions (that use `halfArrowSw`) to override the value (mainly for use outside `arrows.ptl`)
	- Export `TrigArrowShape` for `⯽`
- Exported `AdviceStrokeInSpace` for more fine-grained stroke width calculation
- Modified the metrics of `⊶⊷⧟⟜⊸` so that the calculation is based on the outer radius of the rings
- `⯺⯻⯼⯽` (Chess Pawn notations)
	- Based on the existing mapping symbols `⊶⊷⧟`. These symbols are also modified so that the connecting bar is more visible.
- `⚭⚮⚯` (Genealogy symbols)
	- Implemented as compositions of a "slightly small white circle"
	- Unicode puts `⚬` under the same set as these symbols, but it is too big to compose to these symbols. Used a slightly smaller size instead.
- `𝄬𝄭𝄮𝄯𝄰𝄱` (Accidentals with Up/Down arrow)
	- `♭♮♯` with Up/Down arrow shapes added.
	- On that topic, `𝄲𝄳` is supposed to be a "pitch modifier" different from the quartertone accidentals according to [a public comment in Unicode](https://www.unicode.org/L2/L2024/24223-pubrev.html). I didn't move these glyphs as the encoding of the actual quartertone sharp/flats are not stabilized (as of writing), but just as a FYI.


packages/font-glyphs/src/symbol/geometric/masked.ptl
```diff
@@ -449,3 +449,28 @@ glyph-block Symbol-Geometric-Masked : for-width-kinds WideWidth1
 			include [refer-glyph : MangleName 'whiteTriangleUp'] AS_BASE ALSO_METRICS
 			include : HSerif.lb (Geom.MidX - triangleSideRadius) Geom.Bot jut GeometryStroke
 			include : HSerif.rb (Geom.MidX + triangleSideRadius) Geom.Bot jut GeometryStroke
+
+		create-glyph [MangleName 'marriageSymbol'] [MangleUnicode 0x26AD] : glyph-proc
+			set-width Geom.Width
+			local offsetX : Geom.Size * Size.SlightlySmall.size / 2
+			include : union
+				with-transform [Translate offsetX 0] : refer-glyph : MangleName 'whiteSlightlySmallCircle'
```
be5invis, line 457: Need to make the ring thinner at Heavy.

be5invis (changes requested):

jmcwilliams403:
>  * On that topic, `𝄲𝄳` is supposed to be a "pitch modifier" different from the quartertone accidentals according to [a public comment in Unicode](https://www.unicode.org/L2/L2024/24223-pubrev.html). I didn't move these glyphs as the encoding of the actual quartertone sharp/flats are not stabilized (as of writing), but just as a FYI.

I actually added the characters before the original Unicode proposal for the dedicated characters was even filed.

Logo121:
> > * On that topic, `𝄲𝄳` is supposed to be a "pitch modifier" different from the quartertone accidentals according to [a public comment in Unicode](https://www.unicode.org/L2/L2024/24223-pubrev.html). I didn't move these glyphs as the encoding of the actual quartertone sharp/flats are not stabilized (as of writing), but just as a FYI.
> 
> I actually added the characters to the font before that Unicode proposal for the dedicated characters was even filed.

Yeah I know, just noting a move is needed eventually because they are now (planned to be) encoded as a separate character.

packages/font-glyphs/src/symbol/geometric/masked.ptl
Logo121: Since all the partnership symbols refer to one symbol, I'd prefer not to make a separate ring just for `⚭`. Instead I'm just spacing the gaps more evenly for now.

If this doesn't solve it, it's probably better to use `⚬` directly for `⚭`, at the expense of making it differently sized from `⚮⚯`.

Issue #2603, opened 2024-11-30
U+276F Extended version looks too big


atauzki:
![image](https://github.com/user-attachments/assets/25954c8f-416b-4c75-8d84-7a06c09b697f)

---

* Your font version: 32.2.1
* Your font variant: Fixed
* Your operating system (name and version): Windows 11 26100
* Your application using Iosevka: Windows Terminal

be5invis:
About height or width?

atauzki:
mainly height

atauzki:
the height

jmcwilliams403:
FWIW one of the pairs of dingbat ornamental angle "brackets" (`❮`/`❯`) is actually supposed to be based on guillemets (`‹`/`›`) not angle brackets (`⟨`/`⟩`) hence its different name in Unicode than the other two pairs (`❬`/`❭`/`❰`/`❱`).

be5invis:
Resolved in #2623. This symbol now follows Guillmets' dimension.

real-or-random:
> FWIW one of the pairs of dingbat ornamental angle "brackets" (`❮`/`❯`) is actually supposed to be based on guillemets (`‹`/`›`) not angle brackets (`⟨`/`⟩`) hence its different name in Unicode than the other two pairs (`❬`/`❭`/`❰`/`❱`).

That's certainly true, but given that these are dingbats and in particular ornaments, that doesn't mean that the dimensions need to be the same. And most fonts seem to make them tall. See for example page 4, top left corner, of the [Unicode spec PDF](https://www.unicode.org/charts/PDF/U2700.pdf) for a direct comparison with the normal quotation marks. (I'm aware that the shapes are not normative) 

These ornaments are often used as a terminal prompt (as in the screenshot in the opening comment) because they are tall in most fonts, e.g., the tide prompt for the fish shell uses these.

So this fix broke/changed my prompt, and I assume this will affect many many users.

Sure, I can change my prompt to use a different character (and I will), but since there is no clear right or wrong here, I believe reverting this change will be a good thing for most users.

Ved-un:
Don't do that, fix it back to the way it was, please.

This has disfigured the prompt in the terminal for many people.

Logo121:
It's probably not my call to make on how the glyph should be rescaled, but meanwhile, if you want the original shape back (and have been building your own configurations anyway), you can change these 2 lines in the source code
[...]

PR #2623, opened 2024-12-25
Around 25 star/florette dingbats

Logo121:

`❮❯‹›`
image

```
✛✜✲✢✻✼✽❃✾✿❀❁❊❋
✣✤❉✥❄❅❆❇❈✠❂
```
imageimage

</details>

----

- Made a "generalized `stars.ptl`" for other asterisk/florette dingbats with rotational symmetry. It basically takes a glyph-proc function for 1 petal, then calls it several times `PointingAt` different angles.
- The dingbats be grouped by their shapes:
	- `✛✜✲` (Open Centre)
	- `✢✻✼✽❊❋` (Teardrop-spoked)
	- `✣✤❉` (Balloon-spoked)
	- `✥` (Club-spoked)
	- `❄❅❆` (Snowflakes ~~right in time for Christmas~~)
	- Combinations of shapes:
		- `❇❈` (Sparkles), combining Open Center and Teardrop
		- `❃✾✿❀❁` (Florettes), combining different variants of Teardrop
- Also implemented:
	- `❂`, but as composition in `masked.ptl`.
	- `✠`, but in `cross.ptl`.
- Some implementation notes:
	- The names are not followed exactly. 
		- These characters vary quite a bit between fonts, so I've no idea if the slightly different terms really correspond to a specific feature.
		- For simplicity I just try to make each of the above groups follow a consistent design.
	- "Heavy" asterisks are usually implemented as with a bold stroke (or something less heavy), since it just wouldn't fit into the glyph space.
	- Because the rotational glyphs all use `PointingAt`, they do not follow the stroke contrast unlike the existing asterisks. It's probably too complicated to apply stroke contrast for most of the shapes.
- Reduced height/width of `❮❯` to match angle quotation marks (#2603)


packages/font-glyphs/src/symbol/geometric/rotational.ptl
```diff
@@ -0,0 +1,275 @@
...
+			include : union.apply null : for-rotational-pointing 4 0 Geom.Size 0 : Club normalSw (ballRad * 1.35)
+
+		# Snowflake
+		create-glyph [MangleName 'snowflake'] [MangleUnicode 0x2744] : glyph-proc
+			set-width Geom.Width
+			include : union.apply null : for-rotational-pointing 6 0 Geom.Size 0 : Snowflake normalSw normalSw 6 0
+		create-glyph [MangleName 'snowflakeTight'] [MangleUnicode 0x2745] : glyph-proc
+			set-width Geom.Width
+			include : union.apply null : for-rotational-pointing 6 0 Geom.Size 0 : Snowflake normalSw normalSw 6 1
+		create-glyph [MangleName 'snowflakeHeavyChevron'] [MangleUnicode 0x2746] : glyph-proc
```
be5invis, line 240: `snowflakeHeavyChevron` is too thick per the preview image. Consider reduce the stroke width of both the center stems and the chevron, while keeping the chevron being thicker than the center stem.

be5invis (changes requested):

Logo121:
image

Bumped down the thickness of both the stem and chevron by 1 grade. Is this good enough or should it be even thinner?

PR #2651, opened 2025-01-21
Zigzag Shape fixes

Logo121:

`ծ‹›⧘⧙⧚⧛⦚𜺰`, in max slant (`‹›` is not changed)
image

```
┌𜰉𜰉─𜰊𜰊
𜰊───┘𜰊
```
image

</details>

---

This PR mainly tries to address the problem of mask-based composition of line segments, which has these little spikes when slanted:
image
The changes aren't really applicable to the guillemets, which probably needs some special treatment. (Might open a separate issue for this)

- Add a new knot interpolator `sharp-corner`, which simulates a "corner" biknot by calculating the intersection of the stroke edges of 2 straight lines intersecting at some reference point.
	- Assumes:
		- The knot types before/after the reference point leads to a straight line into/out of the reference (e.g. `flat/curl`)
		- The widths for the knots before/after are specified (the stroke information is not given otherwise)
			- The widths for the knots are assumed to be constant for the previous->reference and reference->next stroke segments.
		- The knot sink(?) is dispiro (the sink information is not given)
			- Currently the biknot is just simulated by the midpoint of the 2 edges (+ an absurd `heading` vector)
			- It's technically possible to arrange the knots such that the resulting knots work for both `spiro-outline` and `dispiro`, but it will further complicate the code.
- Add 2 new shared shapes: `Polyline` and `ClosedPolyline`, which returns a list of knots to draw a polyline and polygonal stroke.
	- `ClosedPolyline` is currently not used but is tested to be functional.
- Rewrite `HZigzag` and `VZigzag` to:
	- Include a new variant where the wave starts from the center (for `𜰉𜰊`)
	- Extend the terminals so that the stroke actually reaches the edge (for `𜺰`)
- Use the new functions to:
	- Rewrite `ծ` so that the sharp bend is a bit cleaner in both code and the resulting geometry.
	- Clean up code for `𜺰`, which previously used 2 extra strokes to extend the zigzag.
	- Clean up code for `𜰉𜰊`, which finally uses `HZigzag` `VZigzag` in a relatively clean way.
- Other changes:
	- Split some of the more geometric legacy symbols (including `𜺰`) into its own module `pictograph/geometric.ptl`.

### DIFF
--- a/changes/32.5.0.md
+++ b/changes/32.5.0.md
@@ -1,2 +1,12 @@
 * Add variant selector for decorative angle brackets (U+276C...U+2771) (#2603, #2623).
+* Refine shape of:
+  - ARMENIAN SMALL LETTER CA (`U+056E`).
+  - VERTICAL ZIGZAG LINE (`U+299A`).
+  - LEFT WIGGLY FENCE (`U+29D8`).
+  - RIGHT WIGGLY FENCE (`U+29D9`).
+  - LEFT DOUBLE WIGGLY FENCE (`U+29DA`).
+  - RIGHT DOUBLE WIGGLY FENCE (`U+29DB`).
+  - HORIZONTAL RESISTOR SEGMENT (`U+1CC09`).
+  - VERTICAL RESISTOR SEGMENT (`U+1CC0A`).
+  - HORIZONTAL ZIGZAG LINE (`U+1CEB0`).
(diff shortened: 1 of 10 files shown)


be5invis:
Wow I always wanted to do this but never realized. Good job.

PR #2657, opened 2025-01-25
Rename some of the properties, to make them more intuitive

be5invis:


jmcwilliams403:
I would like to point out that `[DivFrame].advs` could possibly be misunderstood as being related to `AdviceStroke` instead of being a shortening of `AdvanceScale`{`MM`|`M`|`T`|`F`|`I`|`II`}

be5invis:
I don't think we ever abbrivated `adviceStroke` so this should be fine.

PR #2673, opened 2025-02-06
Reduce word spacing in quasi proportional mode

### LINKED ISSUE #2656: Tighten word-spacing / whitespace in quasi proportional mode
The current word spacing in quasi proportional mode is pretty wide. This can make it harder to read. Proportional fonts usually have narrower word spacing. Since monospace constraints don't apply in quasi proportional mode, I suggest reducing the whitespace width, either by default, or by allowing users to specify a variant. 

I created a small demo to show the difference: https://jul-sh.github.io/demo-word-spacing-iosevka/.

I currently [post process my own version of iosevka reduce whitespace width](https://github.com/jul-sh/iosevka-julsh/blob/main/scripts/build_fonts.py). It'd be great to upstream the option.

jul-sh:
Currently Iosevka has word spacing that's quite a bit wider than many other quasi-proportional fonts. This change adjusts that slightly, imo resulting in easier to read text.

I created a small preview site to demo the difference here: https://jul-sh.github.io/demo-word-spacing-iosevka/

Fixes #2656


packages/font-glyphs/src/space/index.ptl
```diff
@@ -14,7 +14,7 @@ glyph-block Spaces : begin
 		set currentGlyph.glyphRank   (9998)
 
 	create-glyph 'space' ' ' : glyph-proc
-		local df : include : DivFrame para.advanceScaleI
+		local df : include : DivFrame para.advanceScaleII
```
be5invis, line 17: I once did some experiments and found that `advanceScaleII` is too narrow. We could add a new advance scale gear specially for spaces.

Edit `parameters.toml` to set it.

be5invis (changes requested):

packages/font-glyphs/src/space/index.ptl
jul-sh: Done! I also made that param a valid metric override.

be5invis:
Could you make a preview? Not sure if 5/9 works better or 7/12 works better.
Aile and Etoile's unit width is 600emu, thus using 7/12 could result in a integer advance width while 5/9 won't.

be5invis:
Also please note your changes into `changes/32.5.1.md` and rename it to `changes/32.6.0.md`.

jul-sh:
done; updated the aforementioned demo to show all three width. I personally like 7/12, made it the current state of the PR.

Issue #2630, opened 2025-01-01
Selectable {O|Oval|Diamond}-shaped Greek Lower Theta (`θ`), like that of digit zero (`0`).


jmcwilliams403:
- [x] The requested variant shape **does not go too far away from Iosevka's design**.
- [x] The requested variant **does not conflict with any characters in Unicode that Iosevka currently supports**.
- [x] For each variant you requested, there are at least **two monospace/programming fonts created by different designers supported the requested variant**. Provide images below.

------

`Oθ0`
Example of the `O` shape:
Jetbrains mono:
![image](https://github.com/user-attachments/assets/480b8fc2-154d-4599-938c-47bf9e006bc2)
Input mono:
![image](https://github.com/user-attachments/assets/5d67cc40-8a8a-4487-b58f-2081b9fcb397)
Example of the diamond shape:
PragmataPro:
![image](https://github.com/user-attachments/assets/5460734e-9dd9-404b-9763-ad7277d12da0)

Issue #1943, opened 2023-08-20
Heavy ASCII/Latin-1 Operators


jmcwilliams403:
- The requested characters are ...
  - [ ] Latin
  - [ ] Cyrillic
  - [ ] Greek
  - [x] Punctuation
  - [ ] Diacritic / Mark
  - [x] Symbol
- The requested characters are used in ...
  - [ ] The alphabet / character set of a natural language.
  - [ ] The alphabet / character set of a constructed language.
  - [ ] The alphabet / character set of a transliteration or notation system.
  - [ ] Mathematical or other professional use in documents.
  - [ ] The symbol set of a programming language. For the languages that support using arbitrary symbols, please provide usages in well-known components or libraries.
  - [ ] The symbol set of a command-line program.
- [x] Some other monospace/programming fonts supported this character. Provide images below.

------

`U+2753`, `U+2757`, `U+2795`..`U+2797`, `U+1F7F0`.

`❓❗➕➖➗🟰`

Julia Mono:
![image](https://github.com/be5invis/Iosevka/assets/37010132/ebb08bed-9aa9-4c5b-9de1-e8d40c811de0)
Fairfax HD:
![image](https://github.com/be5invis/Iosevka/assets/37010132/3b0c6285-72b4-4358-9850-746e7a6cd531)

Additionally there is also `💲` (`U+1F4B2`) shown here in the Unicode sheets for lack of a better example:
![image](https://github.com/be5invis/Iosevka/assets/37010132/eba0f72d-1bdc-450d-adc7-18ed44bfd8da)

be5invis:
These are mostly used as Emojis I think? Are there any active use them in PLs or technical documents?

jmcwilliams403:
The heavy `❓`/`❗`/`➕`/`➖`/`➗` predate emoji, and in fact the latter three (as well as `🟰`) all go with the already-supported heavy `✖` (`U+2716`).
the heavy `🙴` isn't included in the emoji spec and has no standardized emoji variant.
The only one of this that could be considered emoji-first is the heavy `💲`.

There exists a Swift tutorial about Unicode operator overloading using multiple of these heavy operators at [Medium](https://medium.com/codex/swift-create-your-own-custom-operator-a6fe4d71f606), however the full article requires an account.

jmcwilliams403:
[...]

Issue #1727, opened 2023-05-16
[Character variant request] New cool `0` variant from TX-24 Houston Mono from Berkeley Graphics


haunt98:
- [x] The requested variant shape does not go too far away from Iosevka's design.
- [x] The requested variant does not conflict with any characters in Unicode that Iosevka currently supports.
- [ ] At least two monospace/programming fonts, created by different designers, supported the requested variant. Provide images below.

------

Come across TX-24 Houston Mono from Berkeley Graphics in Twitter and I saw their `0` variant.
It's look good! I wish Iosevka would support this variant.

<img width="599" alt="image" src="https://github.com/be5invis/Iosevka/assets/5424272/c7ca8f5b-b1f9-4bd4-aeb7-6bc25f2e8f7f">

[Source](https://twitter.com/berkeleygfx/status/1643217682874130432)

be5invis:
Square or diamond?

haunt98:
Hi @be5invis, I think diamond is way cooler 💯

Logo121:
For the diamond shape, I *guess* PragmataPro does something similar? Basically the OCR-A approach to differentiating O and 0 (though OCR-A uses this shape for O)

arpangreat:
I'm sorry to ask this here , but anyhow can we use cv77 or diamond shape 0 render in terminal?

Roundlay:
Sorry to ping a closed issue. Can anyone detail how to revert to the original rounded `0` with the latest version of the font?

arpangreat:
@Roundlay If you are using any terminal that supports harfbuzz eg kitty or wezterm , you can use font features and set cv77=(to your liking) , here is all of them: https://github.com/be5invis/Iosevka/blob/main/doc/character-variants.md

PR #2686, opened 2025-02-20
4 Smalltalk symbols (and 2 other arrows)

Logo121:
<details> <summary> Preview</summary>

`⏿𜺱𜺳ɱ𜺳🢲🣀🣁`
image

</details>

----

Other changes:
- Made a new module `o-shape-variants.ptl` for some common/shared "ring-like" shapes (kinda in preparation for #2630), i.e.:
	- Oval Shape and Diamond Shape, moved from `0.ptl`
	- Rounded Rectangle Shape (ref: the unrequested `0` shape in #1727), moved from `split-graphic.ptl` (`𜲲𜲳𜲴𜲵`)
- Some code are modified to import from the above module.
- Fix on the logic of `[MaskLeftLine]` and `[MaskRightLine]`.
- Minor substitutions of `MosaicWidth / 2` --> `MosaicMiddle`.

Character additions:
- `⏿`
	- This glyph has a surprising amount of variations among fonts, but this follows PragmataPro's for simplicity.
	- Composition of `<` with an ellipse.
- `🢲`
	- Directly implementable with HookArrow.
- `🣀🣁`
	- Composition+Masking with `←→↓`.
	- The arrows are moved inwards because by default the arrow heads will overlap in WWID.
- `𜺱`
	- Composition with triangle+circle. Implemented as non-mosaic (align with pictographs), but this can be changed if necessary.
- `𜺳 ` (Lower caret)
	- A zero-advance character with a `HalfRectTriangle` shape, essentially pointing between the current and next character.
	- The exact metrics may need tuning.

----
After this the characters left for full Smalltalk charset support would be:
- Heavy `!` (`❗` U+2757, Part of #1943)
- Heavy wedge-tailed rightwards arrow (`➽`, U+27BD)
- Old PC with portrait monitor (`𜺲`, U+1CEB2)
	- tbh I've no idea what the reference glyph is supposed to be/look like (unless this 
	image is somehow acceptable), so I didn't include this implementation into this pr.


dscorbett:
For U+1CEB2 OLD PERSONAL COMPUTER WITH MONITOR IN PORTRAIT ORIENTATION, [the Unicode proposal L2/21-234](https://www.unicode.org/L2/L2021/21234-terminals-smalltalk.pdf) has examples of its glyph in the original font. The glyph depicts [a Xerox Alto](https://www.righto.com/2017/10/the-xerox-alto-smalltalk-and-rewriting.html), an old personal computer that ran Smalltalk and had a portrait-oriented monitor.

Issue #2148, opened 2024-01-03
Partial Derivative (`VSAS`) with a closed contour.


jmcwilliams403:
- [x] The requested variant shape does not go too far away from Iosevka's design.
- [x] The requested variant does not conflict with any characters in Unicode that Iosevka currently supports.
- [x] At least two monospace/programming fonts, created by different designers, supported the requested variant. Provide images below.

------

Disclaimer: This variant should _only_ be available to Partial Derivative (`∂`). It would be inappropriate for Icelandic Eth (`ð`).

`6ð∂`

Andale Mono:
![image](https://github.com/be5invis/Iosevka/assets/37010132/e96c839d-563e-4254-ab84-446dd1eb9dcf)

Consolas:
![image](https://github.com/be5invis/Iosevka/assets/37010132/6cf7989d-0904-495c-8576-89d6314c30ef)

Menlo:
![image](https://github.com/be5invis/Iosevka/assets/37010132/54106a59-b634-4a4f-bb66-e8586c1c748a)

Liberation Mono:
![image](https://github.com/be5invis/Iosevka/assets/37010132/9f1d641c-8942-4d7f-b348-7be1e9c5948f)

Pragmata Pro:
![image](https://github.com/be5invis/Iosevka/assets/37010132/28168d3a-a50d-410b-9fa1-fc6111a4ad68)

Source Code Pro:
![image](https://github.com/be5invis/Iosevka/assets/37010132/f850b7c7-4468-410b-aede-5bbfb5113f43)

Ubuntu Mono:
![image](https://github.com/be5invis/Iosevka/assets/37010132/b3bbf9ac-b5c6-4b90-a176-026c1bfdff00)

Lucida Console:
![image](https://github.com/be5invis/Iosevka/assets/37010132/fa30c090-781a-481b-99cc-edbf7cb56f0d)

elu00:
Is this effectively adding a variant to revert #862? If so, I'd definitely appreciate it as well!

PR #2687, opened 2025-02-21
Theta and Partial Derivative variants (#2630, #2148)

Logo121:
image

- Add variant selector (`cv71`) for the different body shape thetas (#2630)
	- It kinda ended up using the same `ZeroShapeBase` code for the body/mask construction. May need cleanup.
- Add variant for the old closed contour partial derivative (#2148)
- Cleanup on imports in `o-shape-variants.ptl`.


jmcwilliams403:
You forgot to remove instances of `nonBreakingTagForNewVariantSelector` (which fwiw we haven't been properly removing in the past couple of full-version bumps).

params/variants.toml
```diff
@@ -8821,6 +8840,7 @@ capital-gamma = "serifless"
 lower-gamma = "casual"
 capital-delta = "straight"
 lower-delta = "rounded"
+lower-theta = "oval"
 lower-iota = "serifed-semi-tailed"
 capital-lambda = "straight-serifless"
 lower-lambda = "straight"
```
be5invis, line 8846: Please update the stylistic sets' assignments also (like PragmataPro used diamond-shaped theta).

be5invis (changes requested):

params/variants.toml
Logo121: updated at least the ones mentioned in the corresponding issues for now.

PR #2688, opened 2025-02-23
Move `partial-derivative` variant selector tag to the end of Greek section.

jmcwilliams403:
This basically treats it like an equivalent to e.g. _var-delta_ like how it's treated in many mathematical alphanumeric character sets (not just the Unicode block but also some (La)TeX implementations as well as multiple Symbolics Lisp terminal keyboards).


be5invis:
Make sure you correctly noted the tag changes in `changes/` file.

jmcwilliams403:
> Make sure you correctly noted the tag changes in `changes/` file.

The doc for character variant tags' changes so far was already outdated upon #2687 which originally changed them in the first place.

Issue #2696, opened 2025-02-27
Alternative serif shape for K and k (and Cyrillic К, к, Ж, and ж).


van-de-bugger:
- [x] The requested variant shape **does not go too far away from Iosevka's design**.
- [x] The requested variant **does not conflict with any characters in Unicode that Iosevka currently supports**.
- [x] For each variant you requested, there are at least **two monospace/programming fonts created by different designers supported the requested variant**. Provide images below.

------

Current Iosevka Slab `K` and `k` serif shape (red areas do not look like serifs, they look like main stroke turns):

![Image](https://github.com/user-attachments/assets/5fcd24f5-7768-41e2-baf3-ff91a92c27ea)

Requested serif shape:

![Image](https://github.com/user-attachments/assets/e05bd803-3195-4caf-ac6b-c4c27974395d)

**Monospace fonts with the requested serif shape are:**

APL2741 (by Adrian Smith):
![Image](https://github.com/user-attachments/assets/331fafbb-4592-479d-baf6-93b7c51287c8)
Average mono (by Richard Alexander Hall):
![Image](https://github.com/user-attachments/assets/7d43b79d-4217-4be2-9510-717fa4fafc74)
Computer Modern Typewriter (by Donald E. Knuth):
![Image](https://github.com/user-attachments/assets/5a6f5424-6da2-45b0-98c2-ab0902c69c6e)
Courier Prime (by Alan Dague-Greene):
![Image](https://github.com/user-attachments/assets/87ada078-31a2-4ffd-92b9-58b96dfdb07b)
Cutive Mono (by Vernon Adams):
![Image](https://github.com/user-attachments/assets/d456883b-00b0-4cc0-89e4-4ac84a64b969)
GNU Freefont Mono (by Primož Peterlin, Steve White):
![Image](https://github.com/user-attachments/assets/58d7d0d6-3a0b-4a86-b08e-51386f74c691)
Go Mono (by Bigelow & Holmes):
![Image](https://github.com/user-attachments/assets/b3ce6882-21dd-4189-ba73-2b761e515fc4)
Latin Mono (by Donald E. Knuth):
![Image](https://github.com/user-attachments/assets/cc3d2cb5-4d70-49b4-a028-e2152e74d606)
Luxi Mono (by Bigelow & Holmes):
![Image](https://github.com/user-attachments/assets/0674a7f6-461a-4623-94b2-ff0688ee8db2)
Monaspace Xenon:
[...]

PR #2700, opened 2025-03-03
Add full-serifed variants for `K` and `k`, and related letters (#2696).

be5invis:


packages/font-glyphs/src/letter/latin/k.ptl
```diff
@@ -297,11 +358,18 @@ glyph-block Letter-Latin-K : begin
...
 
-			include : difference
-				union
-					KSlabs        [if slabLegs 1 0] top left right stroke true
-					KAttachment attachment slabLegs top left right stroke true
-				HalfRectTriangle (dim.kshRight - dim.Ok - TINY) 0 (dim.arcTerminalX - TINY) dim.arcTerminalY
+			define [BotSerifMask] : HalfRectTriangle (dim.kshRight - dim.Ok - TINY) 0 (dim.arcTerminalX - TINY) dim.arcTerminalY
+
+			include : KLegSlabs
+				mode        -- [if slabLegs 1 0]
```
jmcwilliams403, line 364: Possibly due to this line, Cursive `k` without a diagonal tail but with full serifs is exactly the same as all that but with non-full serifs.

jmcwilliams403, line 364: Possibly due to this line, Cursive `k` without a diagonal tail but with full serifs is exactly the same as all that but with non-full serifs.

PR #2710, opened 2025-03-17
Flat top variants of `Ww`

### LINKED ISSUE #2146: Curly `W`/`w` with a flat top.

Logo121:
Closes #2146. This is separated from #2707 due to some problems found last minute.

<details> <summary> <code>wWⱳⱲ</code> with <code>cv57/cv32 = 3, 6, ..., 33</code>  </summary>

image

</details>

- Added a separate enumeration(?) `midHClass` for middle stroke height.
	- It _might_ be possible to make this a separate variant dimension, but I'll let you decide first.
- The implementation is a bit messy -- initially I misunderstood the variant to be moving the middle connection up to the top, instead of having an extra middle stem.
	- A new variable/dimension `wMiddleHeightExt` is added for the distinction between the connection and terminal y-position (needed for Curly and VerticalSides)
	- The logic behind the 2 height functions:
		- For `FORM-CURLY` and `FORM-VERTICAL`, `wMiddleHeight` has its specific values.
		- For the other forms, `wMiddleHeight` references `wMiddleHeightExt` for possible override, but defaults to:
			- Some specific values for `FORM-STRAIGHT` and `FORM-CURSIVE`
			- `top` for the rest.
		- `wMiddleHeightExt` depends on `midHClass` (and `wMiddleHeight` as a fallback)
	- Might need some cleanup/refactoring if possible.
- The full-serif top-aligned forms now all use outward serifs like the double-v forms. This is to prevent stroke overlaps.
	- Might not be needed for QP/wide forms? This is not handled in this PR.
- These variants are in theory implementable but are not included as there are no existing requests:
	- `CURSIVE` forms with extended middle stem
	- `ALMOST-TOP` forms for all of the above
	- `VERTICAL` forms with the middle connection aligned to the top instead of a stem extension, like the corresponding `M` variants.
- Other changelog changes:
	- Bumped to `33.2.0` because of new variant.
	- I accidentally used <code>`'</code> in the previous changelog. Might be too late to change it because 33.1.0 is already released, but i'll fix it anyway
		- (There's actually a much older mistake in `22.1.1.md` where I wrote "little" in place of "small" letter. I would like to fix that too if that's allowed.)
	- Since v33.1.0 didn't include #2708, the variant assignment fix is noted as well.

### DIFF
--- a/packages/font-glyphs/src/letter/latin/w.ptl
+++ b/packages/font-glyphs/src/letter/latin/w.ptl
@@ -15,11 +15,15 @@ glyph-block Letter-Latin-W : begin
 
 	define FORM-CURLY             0
 	define FORM-STRAIGHT          1
-	define FORM-ALMOST-FLAT-TOP   2
-	define FORM-FLAT-TOP          3
-	define FORM-DOUBLE-V          4
-	define FORM-ASYMMETRIC        5
+	define FORM-VERTICAL          2
+	define FORM-DOUBLE-V          3
+	define FORM-ASYMMETRIC        4
+	define FORM-CURSIVE           5
 	define FORM-CYRL-OMEGA        6
+
+	define MIDH-OTHER             0
+	define MIDH-TOP               1
+	define MIDH-ALMOST-TOP        2
 
 	define SERIFS-NONE            0
 	define SERIFS-AUTO            1
@@ -29,8 +33,8 @@ glyph-block Letter-Latin-W : begin
 	define SERIFS-CURSIVE         5
 	define SERIFS-CYRL-OMEGA      6
 	define SERIFS-ALL             7
-	define SERIFS-DOUBLE-V-ALL    8
-	define SERIFS-ASYMMETRIC-ALL  9
+	define SERIFS-ALL-OUTER       8
+	define SERIFS-ALL-ASYMMETRIC  9
 
 	define CwFineOuter            5.0
 	define CwFineInner            6.0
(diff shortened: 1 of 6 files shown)


jmcwilliams403:
You forgot to add `nonBreakingVariantAdditionPriority = 100` for the new variants.

jmcwilliams403:
I don't think the outer serifs are correctly implemented for vertical-sides variants.
Notice here how they're detached under thin and have a longer jut length than the symmetric serifs:
image

be5invis:
LGTM

packages/font-glyphs/src/letter/latin/w.ptl
```diff
@@ -135,11 +135,37 @@ glyph-block Letter-Latin-W : begin
...
 
-		return : object xBot1 xBot2 strokeOuter strokeOuterCr fineOuter fineOuterCr fineInner fineInnerCr fineInnerCrTop fineHeight wMidHeight xRight1 xLeft2 wCoCr curlyStraightSegLength
+		return : object xBot1 xBot2 strokeOuter strokeOuterCr fineOuter fineOuterCr fineInner fineInnerCr fineInnerCrTop fineHeight wMidHeight wMidHeightExt xRight1 xLeft2 wCoCr curlyStraightSegLength
+
+	define [WSerifs df top bodyType slabType dim] : glyph-proc
+		local oSlabPos : match bodyType
+			[Just FORM-CURLY]      : O * 2
+			[Just FORM-VERTICAL]   : O * 2
+			[Just FORM-CYRL-OMEGA] : O * 2
+			__                       0
```
Logo121, line 145: looks like I reversed the logic here. `O * 2` and `0` should be swapped with each other.

since this is merged already i'll just leave this here for cleanup reference. I have nothing else to push for now.

PR #2713, opened 2025-03-19
Make Lower `w` use `straight-flat-top-serifless`/`straight-almost-flat-top-serifed` by default.

jmcwilliams403:
I made this a separate PR from #2712 so you can have the ability to accept/reject it separately if you want.

I've wanted to do this for ages but it wouldn't have made sense to do before without `curly-flat-top` existing first (for `ss20`).

This only affects the lowercase. The capital is unchanged.

This also uses cursive forms for lower `v`/`w` under Slab Italic to match `k`/`x`/`z` and `ss15`/`ss17`.

```
```

Sans Upright:
image
Sans Italic:
image
Slab Upright:
image
Slab Italic:
image


be5invis:
Sorry, I do not want to change the default style for now.

Resolve as by design.

PR #2763, opened 2025-05-09
`arrow.ptl` Refactor

Logo121:
Still needs cleanup and checking for errors.

This PR tries to break down `arrow.ptl` into separate groups within the `arrow` folder:
- Simple: Everything that can be constructed with a single `MkArrow` call
- (Arrows with small extensions):
	- Dashed: Dashed arrows
	- Angled: Angled/Bent arrows
	- Hook-Loop: Hook and Loop-tail arrows
	- Circular: Arrows with a `OShape`-based bar 
	- Zigzag: Zigzag and Lightning arrows
	- Wave: Wave/Squiggle arrows
	- Legacy: Legacy Computing arrows (that are not part of the other groups)
- (Arrows with their own set of implementations):
	- Thick: Black or White "thick" arrows
	- Rounded: for that single Rounded arrow
	- Weighted: Arrows with different weight/geometry specifications
- Compositions:
	- Stacked: Arrows implemented using VStack and HStack (mostly)
	- Composite: Other arrows using some composition method

The shared functions are in `arrow/shared.ptl`, which holds a `ArrowDims` function and other namespaces that take in the dims so that they can actually be exported a bit more cleanly.

Other changes:
- Reimplemented `OutMask`. Forgot what the original shape is supposed to be but this makes more sense and is more flexible.
- The gaps used in `⟲⟳` are identified with Barb arrow head (with a negative thickness), since their implementations were equivalent.
- Reducing a few shape implementations so that they can be constructed more simply:
	- Bar/Tail-to-head arrows are now interpreted as a double-sided arrow with a bar/tail "head" on the other side.
	- Arrow-to-bar glyphs are implemented with a "wrapper" over other arrow functions, reducing the need of creating extra glyphs for composition.
- Commented out some arrow glyphs with no codepoint that didn't seemed to be used elsewhere.

Also implemented `⤌⤍⤎⤏` as a quick extension test, adapting code from dashed arrow.


Logo121:
Comparison (mostly to check if the reimplementations messed up anything): 

After:

Before:

Subtracting `Before - After` for better comparison:


- These images are made before merging more recent commits like #2757, so the thickness wasn't changed
- The differences in the whiteBarred arrows are (hopefully) fixed in the latest commit
- The other differences are mostly due to inconsistent implementations of the barred arrows before this PR.

PR #2771, opened 2025-05-18
fix: update release checksum output format in conformance with spec

jsejcksn:
When verifying checksums for release assets, verification fails due to malformed input. For example: using `sha256sum` in zsh on macOS in a directory containing all of the assets from the most recent [release 33.2.3](https://github.com/be5invis/Iosevka/releases/tag/v33.2.3):

```
% sha256sum --check --strict --quiet SHA-256.txt; echo $?
sha256sum: WARNING: 452 lines are improperly formatted
1
```

The GNU coreutils manual [specifies](https://www.gnu.org/software/coreutils/manual/coreutils#cksum-output-modes-1) the following regarding checksum output formatting:

> **Untagged output format**
> With the `--untagged` option and the `--algorithm` option selecting non legacy checksums, the following output format is used. This is the default output format of the standalone checksum utilities. For each _file_, we print the checksum, a space, a flag indicating binary or text input mode, and the file name. Binary mode is indicated with ‘`*`’, text mode with ‘` `’ (space). Binary mode is the default on systems where it’s significant, otherwise text mode is the default.

Currently each line is being generated with a tab between the hex checksum and filename. If this is changed to conform to the spec (replace each tab with two spaces), then verification succeeds:

```
% sed -i '' -e 's/\t/  /g' SHA-256.txt

% sha256sum --check --strict --quiet SHA-256.txt; echo $?
0
```

This PR implements that change at the site of generation.


jsejcksn:
@be5invis Thanks for changing the base branch for me. There was no instruction to choose [`dev`](https://github.com/be5invis/Iosevka/tree/dev) as the base for the PR prior to creation. Are you interested in including that info in a [PR template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository) for this repo? If, so I could make a PR for it.

PR #2819, opened 2025-07-20
Shrink side bearings and increase bowl size of Cyrillic Ef (`Ф`) under Quasi-Proportional.

jmcwilliams403:
Basically making it match the bowl width of e.g. `Р`/etc. more (which that character has a slightly expanded left SB).

I originally tried changing the advance width of the character (for `Ф`), but it looked too extreme and actually made the negative space worse as well. The solution I found was to leave the advance width alone and shrink the side bearings instead.

This only affects Quasi-Proportional; The line containing `df.adws % 0.5` makes it so that this neither affects monospace nor Bulgarian `Ф` (which uses `para.advanceScaleMM`) as they are already divisible by the number of vertical lines (i.e. multiples of `0.5`) and therefore look fine as-is.

Multiplying the resulting value (of the aforementioned expression) by `0.75` also makes the difference more mild (that is to say _less_ extreme), going from 1/3 to just 1/4 of the side bearing shaved off.

I showed the following samples to some people in a few servers and the general consensus (among those familiar with Cyrillic) seemed to be that the wider bowl looked better. It could always stand to be tweaked in the future, though, depending on taste.

For all screenshots below: Top row is before, bottom row is after.

`ΦϕϷФф РФ: Феличан наскиџтагон`

Aile Thin:
image
Aile Regular:
image
Aile Heavy:
image
Etoile Thin:
image
Etoile Regular:
image
Etoile Heavy:
image

### DIFF
--- a/packages/font-glyphs/src/letter/greek/phi.ptl
+++ b/packages/font-glyphs/src/letter/greek/phi.ptl
@@ -13,12 +13,14 @@ glyph-block Letter-Greek-Phi : begin
 
 	define [VarPhiRing fFlatTB df y2 y3 ada adb] : glyph-proc
 		include : VBar.m df.middle y2 y3 df.mvs
+		local l : mix df.leftSB 0 : if fFlatTB (0.75 * (df.adws % 0.5)) 0
+		local r : df.width - l
 		include : if fFlatTB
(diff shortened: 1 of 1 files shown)


jmcwilliams403:
FWIW this does indeed also affect the non-`split` variants of the lowercase (again only under Quasi-Proportional), but it might also be beneficial for the lowercase as well:

`кюфтенца`

image

PR #2820, opened 2025-07-21
Remove vertical overshoot from letters which use `OShapeFlatTB`.

jmcwilliams403:
The vast majority of letters that use `OShapeFlatTB` use it for `Þ`/`D`-like bowls. In fact it, was originally used exclusively for Cyrillic Ef (`Ф`/`ф`, non-`split` variants only) before I expanded it to some roman numerals and Cyrillic Broad On (`Ѻ`/`ѻ`, not shown here).

Speaking of roman numerals, the overshoot also causes hinting problems with `ↂ`/`ↈ` under regular weight, which is shown in the included samples (this PR appears to have fixed it, at least on my end).

For each screenshot below, top row is before, bottom row is after.

**Note: Greek** `Φ`/`ϕ` **is shown for comparison, but is unchanged.**

`Φϕ ϷФ фᴅ РФ Dↀↁↂↇↈ`

Sans Monospace Thin:
image
Sans Monospace Regular:
image
Sans Monospace Heavy:
image
Slab Monospace Thin:
image
Slab Monospace Regular:
image
Slab Monospace Heavy:
image
Aile Thin:
image
Aile Regular:
image
Aile Heavy:
image
Etoile Thin:
image
Etoile Regular:
image
Etoile Heavy:
image


jmcwilliams403:
BTW the "before" row in the samples is the release version, so it's before my previous PR too.

PR #2848, opened 2025-08-28
Finishing touches for double-storey `a`.

jmcwilliams403:
As it turns out, the _entire expression_ `[HSwToV Stroke]` seemed to have been what was serving as the stand-in for `SmoothAdjust`.
Previously, removing `HSwToV` would break the geometry, but now that `YSmoothMidL` has been flipped right­side-up via #2839 , it is no longer needed.

For each screenshot below: Top row is before, bottom row is after.

`aꜳæꜵꜷꙗ𝕒`

Monospace Thin Upright:
image
Monospace Regular Upright:
image
Monospace Heavy Upright:
image
Monospace Thin Italic under `'cv36'3,'cv87'3`:
image
Monospace Regular Italic under `'cv36'3,'cv87'3`:
image
Monospace Heavy Italic under `'cv36'3,'cv87'3`:
image
Aile Thin Upright:
image
Aile Regular Upright:
image
Aile Heavy Upright:
image
Aile Thin Italic:
image
Aile Regular Italic:
image
Aile Heavy Italic:
image

### DIFF
--- a/packages/font-glyphs/src/letter/latin/lower-a.ptl
+++ b/packages/font-glyphs/src/letter/latin/lower-a.ptl
@@ -60,7 +60,7 @@ glyph-block Letter-Latin-Lower-A : begin
 				[if isMask corner flat] (df.rightSB + O) barTop [heading Leftward]
 				curl [mix df.rightSB df.middle 0.9] barTop
 				archv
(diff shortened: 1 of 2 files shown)


jmcwilliams403:
It turns out that `m.ptl` also uses decade-old code for its arch depths, but with its own assumptions that are separately motivated from that of `a.ptl`.

PR #2860, opened 2025-09-10
Adjust the position of serifed-flat-tailed `l`.

1024th:
- Apply `IBalance2` leftward correction to both 'flat-tailed' and 'serifed-flat-tailed' `l`.
- Add `isForI` parameter `XMiddle` functions to distinguish between `i` and `l`, so the character positioning of `i` is unchanged.

The image below demonstrates this change.

lower-l-font-comparison


1024th:
Sorry, I should commit to the dev branch. Close this PR now.

PR #2861, opened 2025-09-10
Adjust the position of serifed-flat-tailed `l`.

1024th:
- Apply `IBalance2` leftward correction to both 'flat-tailed' and 'serifed-flat-tailed' `l`.
- Add `isForI` parameter `XMiddle` functions to distinguish between `i` and `l`, so the character positioning of `i` is unchanged.

The image below demonstrates this change.

lower-l-font-comparison

### DIFF
--- a/packages/font-glyphs/src/letter/latin/lower-il.ptl
+++ b/packages/font-glyphs/src/letter/latin/lower-il.ptl
@@ -21,18 +21,18 @@ glyph-block Letter-Latin-Lower-I : begin
 	define [TailedDotlessIShift df] : (1 - df.adws) * 0.2
 
 	define XMiddle : namespace
-		define [FlatTailedImpl df addTopSerif] : df.middle - [if addTopSerif 0 : IBalance2 df]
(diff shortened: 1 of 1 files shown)


1024th:
I found a `XMiddle.FlatTailed` in the body of the function `Body.DiagTailed`, but I think this should be replaced by the `xMiddle` parameter of this function? Why isn't this triggering a runtime error (because I think the type of `[XMiddle.FlatTailed df]` is a function but it is expected to be a number)?

```
export : define [DiagTailed df top xMiddle] : glyph-proc
	set-base-anchor 'trailing' (xMiddle + [HSwToV : 0.75 * df.mvs]) 0
	set-base-anchor 'palatalHookMask' currentGlyph.baseAnchors.trailing.x (HalfStroke + O)

	local depth : [DiagTail.StdDepth df df.mvs] * (df.rightSB - xMiddle) / (df.rightSB - [XMiddle.FlatTailed df])
...
```

be5invis:
Well this language compiles to JavaScript. Using your way of modifying the code, missing `isForI` parameter will simply set it to `undefined` (equivalent to false).

PR #2871, opened 2025-09-21
Riyal sign (U+20C1)

Logo121:
image

- `⃁` (U+20C1)
	- Rough implementation referencing [the official guidelines](https://www.sama.gov.sa/ar-sa/Currency/SRS/Documents/Guidelines.pdf) (mostly the metrics on a 17:19 grid), trying to keep the proportions while adapting to differing stroke widths
	- These elements from the official design are skipped (unless deemed necessary):
		- Rounded and slanted terminals (except the top cuts)
		- The cut on the Lam stroke: Not sure if it's necessary but it looks fine without it?


be5invis:
I'd like to close this for now, as...

 1. We do not support any Arabic script yet.
 2. We need to know many details about if we could omit these non-important features, even if this symbol can have weight.

Let us wait for more implementations before we add it.

Logo121:
> 1. We do not support any Arabic script yet.

I don't think this is a problem, since we have Isreali shekel without Hebrew support (aside from the 4 letterlikes). This new symbol is designed to be removed enough from actual Arabic typography to "align with global currency symbols" anyway.

For the symbol's tolerance for variations, I kinda just treated it the same way as other currency symbols (I didn't follow the circular design for Turkish Lira, at least). I haven't seen any implementation that is not a carbon copy of the official symbol (and, well, *Unifont*), but the "usage rules" in SAMA's guideline didn't explicitly disallow omitting those features either, at least...

Feel free to close this. I just kinda started this because I saw the same symbol requested in Inter https://github.com/rsms/inter/issues/842 (Inter doesn't have Arabic support either), so I assumed there's probably a need.

be5invis:
I will close it for now but @Logo121 feel free to reopen once there are more implementations and guidelines (like from Noto, etc.).

ps. for reference, the Estimated sign (U+212E) in Iosevka was reimplemented to match its specification. As a result it does not have weight.

Issue #2879, opened 2025-09-29
Change the ф in aile  and etoile back to a round form


TimurGerasimov250:
Split bowl makes it really ugly, it would be nice to have a more simple round form. ф with a single full bowl, without split bowl makes it more modern and clean. So try it, it will make Aile and Etoile better with round full bowl ф instead of split bowl on ф

PR #2880, opened 2025-09-30
Changed the default style of Cyrillic lowercase Ef (ф) in Aile and Etoile (upright) to use non-split rings, for better style consistency (#2879).

be5invis:
#2879


jmcwilliams403:
You forgot about the split ring cursive bar variant used by Etoile under italic.

be5invis:
> You forgot about the split ring cursive bar variant used by Etoile under italic.

This is intentional

TimurGerasimov250:
Also it is something worse in Aile than Sans. So with non-split rings it means everything is better.

TimurGerasimov250:
Non-split, round bowl

PR #2914, opened 2025-10-20
Make serifs of Cyrillic Lower Dwe (`ꚁ`) respond to italics.

jmcwilliams403:
This is in lieu of a proper italic form — which technically existed — until it gets added to the font. If that ever happens, this PR at least serves as minor cleanup.

This also enforces the serif behavior on _all_ (upright) Cyrillic Lower De derived characters, even though it's only ever visible for `ᲁ` (long-legged de) and `ꚁ` (this character) since the rest of them have proper italic forms (or, in the case of Lower Soft De `ꙣ`, covers up the only serif that actually changes with its augmentation).

`Ддᲁ Ꚁꚁ Лл`

Thin:
image
Regular:
image
Heavy:
image

### DIFF
--- a/changes/33.3.4.md
+++ b/changes/33.3.4.md
@@ -1,0 +1,1 @@
+* Refine serifs of Cyrillic Lower Dwe (`U+A681`) under italics.
--- a/packages/font-glyphs/src/letter/cyrillic/de.ptl
+++ b/packages/font-glyphs/src/letter/cyrillic/de.ptl
@@ -12,11 +12,10 @@ glyph-block Letter-Cyrillic-De : begin
 
 	glyph-block-export BottomExtension
 	local BottomExtension : (-LongVJut) + QuarterStroke
-	local TopExtension        LongVJut
 
 	glyph-block-export CyrDeBottom
 	define [CyrDeBottom left right _sw _desc] : glyph-proc
-		local descenderOverflow : if SLAB SideJut ((right - left) * 0.075)
+		local descenderOverflow : if SLAB SideJut : (right - left) * 0.075
 		local xTopLeft  : mix left right : StrokeWidthBlend 0.15 0.1
 		local xTopRight : mix left right : StrokeWidthBlend 0.95 0.96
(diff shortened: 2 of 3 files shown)


be5invis:
@dscorbett Does this letter have true (cursive-ish) italic form?

jmcwilliams403:
> Does this letter have true (cursive-ish) italic form?

[source](https://en.wikipedia.org/wiki/Abkhaz_alphabet#/media/File:Abkhazian_alphabet_Chochua_1909_in_cursive.webp)

image

jmcwilliams403:
Mind you this PR does not add a new italic form, just some motion serif behavior for the upright form when under italics as a somewhat temporary measure.

dscorbett:
That’s the same example glyph I would have responded with, but, for future reference, I’ve never been able to find an example of this Abkhaz alphabet in true italics as opposed to handwritten cursive. They are related, but handwriting is often curlier or otherwise different, so a handwritten example is not necessarily a reliable indicator for the italic form. For example, [<i lang=ru>Абхазский алфавит</i> (1892), pp.&nbsp;13–14](http://apsnyteka.org/file/Machavariani_Abkhazskaya_azbuka.pdf#page=19) gives the full alphabet in “<span lang=ru>Печат.</span>” (print) and “<span lang=ru>Скороп.</span>” (handwriting) and some of the handwritten letters (especially the capital ones) don’t match Iosevka’s italics; in the bilingual Abkhaz–Russian parts, when the Russian text is italicized, the corresponding Abkhaz text is not, so it might be that no italic font was ever cut for this orthography. I did find a non-italic version of that glyph, though, in [<i lang=ru>Из истории письменности в Абхазии</i> (1967), p.&nbsp;43](https://web.archive.org/web/20220701064308/https://www.unicode.org/udhr/n/abk/bgazhba.pdf#page=23), which (if reliable) implies that if an italic small dwe does exist, it probably does indeed look essentially the same as the handwritten one. In summary, this evidence doesn’t change anything, but I hope it was interesting anyway.

PR #2934, opened 2025-11-05
Move `[DiagTail.[DefaultInnerRadius]]` from `letter/shared.ptl` to `meta/aesthetics.ptl`.

jmcwilliams403:
This saves an import in `letter/latin/v.ptl` and `letter/latin/w.ptl`.

Also further code cleanup in the surrounding files.

Effectively nothing is changed otherwise, although there might be some incidental improvements of floating point rounding by simplifying some of the math with some of the additional code cleanup in `letter/latin/v.ptl`, particularly for Lower V with Curl (`ⱴ`).

### DIFF
--- a/packages/font-glyphs/src/letter/latin/k.ptl
+++ b/packages/font-glyphs/src/letter/latin/k.ptl
@@ -307,23 +307,23 @@ glyph-block Letter-Latin-K : begin
 			define arcFine ShoulderFine
 			define arcStroke : AdviceStroke2 2 3 top
 			define arcTerminalFine : AdviceStroke2 3 4 top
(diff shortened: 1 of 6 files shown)


be5invis:
I think isolating `DiagTail`-specific parameters into separate files is good for organizing things.

Actually I somehow want to move some of the parameters _out_ from `meta/aesthetics.ptl`, and put them into the corresponded functions.

jmcwilliams403:
Compromise: How about we only move `[DefaultInnerRadius]` out? The cursive `w`/`v` shapes have little to do with a literal diagonal tail but still use this specific metric.

be5invis:
You can rename it to `DefaltTightBendInnerRadius`

PR #2944, opened 2025-11-10
Add all modified chars to APL variant

slotThe:
At this point, there are quite a few APL (and related languages) chars that have been adjusted, but `variant.toml` (and thus also the generated before/after pictures) only contains the initial set. 

(I think I did that right? If I read the makefile correctly, this should cause the SVGs that are generate with every release to be updated with these symbols)

### DIFF
--- a/params/variants.toml
+++ b/params/variants.toml
@@ -9266,7 +9266,7 @@ snapshotFeatureApplication = { 'onum' = 1 }
 # This is a special variant selector that controls APL form
 [prime.apl-form]
 isSpecial = true
-hotChars = "∆∇∊○←→↑↓"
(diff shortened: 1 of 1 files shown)


be5invis:
hotChars is only for the samples in the website / document.
No need to include all of them.

slotThe:
Is there any other place where this would fit? Right now one has to dig into the source code in order to find all affected glyphs

Logo121:
Possibly `cv-influences.md`? (seems to be auto-generated, see https://github.com/be5invis/Iosevka/commit/e7368f16e7c9a0ab9e14a756955a2ac817aa3e0a)

Right now it only documents CV tag influences, but it can probably be extended to include non-CV tags as well (APLF, MOSC, NWID/WWID? etc)

be5invis:
I am creating #2950 to replace this -- include in `doc/cv-influences.md`.
Closing.

PR #2947, opened 2025-11-12
Treat diagonal-tailed `ij` variants like hooky (not flat-tailed)

1024th:
This shifts `j` rightward and `i` slightly leftward. The diagonal-tailed `j` in the Aile version is also wider.

Also remove the outdated `dfHook` in `Body.DiagonalTailed` of lower-j. It looks that the correct `DivFrame` is chosen later in `JConfig`.

But the `dfHook` in `Body.FlatHook` of lower-j remains unchanged, because I don't quite understand the code.

### DIFF
--- a/packages/font-glyphs/src/letter/latin/lower-j.ptl
+++ b/packages/font-glyphs/src/letter/latin/lower-j.ptl
@@ -38,6 +38,7 @@ glyph-block Letter-Latin-Lower-J : begin
 		export : define [StraightSerifless df] : begin df.middle
 		export : define [StraightSerifed df]   : df.middle + [IBalance2 df]
 		export : define [BentHook df]          : df.middle + JBalance
+		export : define [DiagonalTailed df]    : df.middle + [IBalance2 df]
 		export : define [FlatHookSerifless df] : df.middle + 0.25 * JBalance * df.adws * [mix 1 df.adws 2]
 		export : define [FlatHookSerifed df]   : df.middle + JBalance * df.adws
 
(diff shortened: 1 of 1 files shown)


be5invis:
`j` may need some alternative spacing settings

packages/font-glyphs/src/letter/latin/lower-j.ptl
```diff
@@ -98,15 +97,15 @@ glyph-block Letter-Latin-Lower-J : begin
 		'bentHookSerifless'          { "BentHook"        Serifs.None   XMiddle.BentHook           Marks.Serifless  Div.BentHook          }
 		'bentHookShortSerifed'       { "BentHook"        Serifs.Short  XMiddle.BentHook           Marks.Serifless  Div.BentHook          }
 		'bentHookSerifed'            { "BentHook"        Serifs.Long   XMiddle.BentHook           Marks.Serifed    Div.BentHook          }
+		'diagonalTailedSerifless'    { "DiagonalTailed"  Serifs.None   XMiddle.BentHook           Marks.Serifless  Div.BentHook          }
```
be5invis: Current space is too large. Need to compress it a little.

be5invis (changes requested):

packages/font-glyphs/src/letter/latin/lower-j.ptl
1024th: What about using the width of `advanceScaleF` and `XMiddle` of `df.middle + [IBalance2 df]` for diagonal-tailed `j`? It looks good to me. Preview:

Issue #2519, opened 2024-09-27
[FR] Sans-serif single-story lower-case a vs o


pallaswept:
- [x] The requested variant shape **does not go too far away from Iosevka's design**.
- [x] The requested variant **does not conflict with any characters in Unicode that Iosevka currently supports**.
- [x] For each variant you requested, there are at least **two monospace/programming fonts created by different designers supported the requested variant**. Provide images below.

------
I am LOVING Iosevka, thanks!! I do find that I am frequently making the same mistake - I mix up my 'a' and 'o'. 

A double-story 'a' is a common means to solve this, but I'm trying to make my Iosevka look like how I write IRL and I don't use double-story unless maybe I'm doing calligraphy. Likewise I don't use tails or serifs, so, that leaves me with few options here. My preferred variant is single-storey-earless-corner-serifless as it matches the other ubuntu-font-style 'corner' variants. I've tried them all, and while they seem OK at large sizes, at smaller sizes like in a terminal, they're near-identical, with only 2 or 3 pixels differing at all.

I'm not sure of the best way to go about resolving this, and I feel like you would have better ideas than any I could imagine, so I haven't provided a specific variant I would like. I am open to suggestion :) One thing that I think might be a good option, is an 'a' where the bottom of the bowl joins the stem higher up. That's common in more cursive fonts, I guess you probably know what I mean, but to help explain myself, here is the first random example I found on some webpage (sorry it's italic): 
![image](https://github.com/user-attachments/assets/db3cd73b-4a44-4039-804e-9e8471806bed)
That's a bit more extreme than I have in mind, but is a good example of what I meant by the bowl joining the stem higher.

This seems like it could maintain the aesthetic of the 'corner' variants, maintain the general bowl shape without needing to change the 'o', and is a very common means to address this problem with distinction between 'a' and 'o'.

Thanks, and I look forward to any ideas you might have about this :)

jmcwilliams403:
Related: #1269

be5invis:
Name it as `a.single-storey-cursive`?

zhenlige:
Maybe changing the bowl to become more asymmetric is also an available choice.
![图片](https://github.com/user-attachments/assets/03523242-7939-450c-93ea-33e0f72606c6)

pallaswept:
Is there anything I could do to assist with this?

pallaswept:
Iasevka is still the best fant 🤣

pok49:
[...]

Issue #1632, opened 2023-03-25
Flat top 'a'


Logo121:
- [x] The requested variant shape does not go too far away from Iosevka's design.
- [x] The requested variant does not conflict with any characters in Unicode that Iosevka currently supports.
- [x] At least two monospace/programming fonts, created by different designers, supported the requested variant. Provide images below.

(unless this shape looking like `Ƌ` is a conflict or something)

------

A somewhat popular variant that is surprisingly not suggested here yet. 
Common for bitmap-like fonts or fonts that imitate that retro-y/scifi-y bitmap style.

Here are some programming fonts with this style (borrowed from https://github.com/braver/programmingfonts)
![image](https://user-images.githubusercontent.com/21302803/227730129-41c4b5ce-2248-4068-bbae-ecc681fdb1f4.png)![image](https://user-images.githubusercontent.com/21302803/227730142-71927c5c-c6ec-4544-9522-85333e8531c8.png)

Issue #1269, opened 2021-12-28
lowercase letter m like  terminus font or ProFont


aplab:
- [ ] The requested variant shape does not go too far away from Iosevka's design.
- [ ] The requested variant does not conflict with any characters in Unicode that Iosevka currently supports.
- [ ] At least two monospace/programming fonts, created by different designers, supported the requested variant. Provide images below.

------

lowercase letter m like in terminus font or ProFont with straight top line

be5invis:
Adding such shapes may also include things like `u` and `n`. Add to the backlog for now.

be5invis:
Ah, it will influence a lot of other letters too, like `b`, `d`, `q`, `p`...
In addition, those "boxy" style doesn't quite match.

aplab:
i mean only additional variant like this
![image](https://user-images.githubusercontent.com/4007240/147606357-13482630-3a33-4ac5-9cac-3085ee56e3ed.png)

be5invis:
Well, when I add a variant I will add a whole suite that influences all the similar letters. Otherwise, other people will throw me thousands of issues.

Martin-Costa129:
I think it's done, why not close the issue?

PR #2957, opened 2025-11-21
Draft: Unified Arc-Hook namespaces

Logo121:
Note: This is a draft from more than a year ago. I suppose this would be relevant now considering the recent PRs on shoulder/bowl/obar like #2930 and #2945, if anyone is interested in continuing this.

No glyphs or glyph functions are dependent on this PR yet (outside PUA `U+E100` to `U+E14F` for demonstration):
image

---
This PR tries to refactor/regularize the Shoulder/Bowl/Hook variant code, organizing them into the 3 recurring styles for different glyph variants ("Toothed", "ToothlessCorner", "ToothlessRounded", + a 4th one), put into a single module `arc-hook.ptl`, and a higher level bowl/shoulder handler in a different module `shoulder-bowl.ptl`. 

The idea is to group the arc and hook functions into namespaces (renamed "Toothed", "Toothless" and "Rounded" tentatively), which provides methods for:
- Arc Start/End, corresponding to the connections of `n` and `u`, for example.
- Hook Start/End, corresponding to the terminals of `c`, for example.
- Hook Serifs, for variants that need them (for those who don't, e.g. Rounded, it returns an empty glyph-proc).
- Y-position for the arc connections, used for specifying the stem boundary for `nudbqp`, for example.
- X-position for the stroke terminals, used for building masks.

The parameters these namespaces take is mostly unified, with some extra params for specific styles:
- ~~Left, Right, Top, Bottom~~ `box`: Bounding box for the position of the arcs
- Stroke, ADA, ADB, Hook
- Fine (only for "Toothed")
- BarSW (`sw` of the connecting stem/serif, for all except "Rounded")
- Rise (only for "Toothless")

This also adds a 4th style to these shapes ("Flat"), relevant for:
- #1269
- #1287
- #1632 (Since the serifed hook variant is implemented as a "Inward" serif)
- Maybe #2519?

which has some extra parameters:
- `slab`: Whether the hook version has a serif or not. No "flat hook with serif" variants are requested but this is kinda just for parity with the other styles that uses `hook`
- `midShift`: Shifting the middle node horizontally away from the terminal side. This is based on the existing implementations of flat arcs/hooks, which never puts the middle node at exactly the middle.
- `xShrink`: Shrinking the terminal node (for hooks) inward. This is based on the existing flat hooks for `gy`, where the terminals don't really reach the boundary.

---
Issues to be resolved (updated after merge):
- Considerations on the specific modifications from existing shapes (whether to incorporate/drop/keep them separate):
	- The `alsoThru` knot in `OBar.toothless` (and thus the `DMBlend` constant)
	- The very specific calculations in double-storey `a`'s compressed bowl knot
	- `cheBowl`, which is a variant of Toothless connection
	- Flat hooks for `gy` (terminals shrink 1/5 of the character width inwards), which is not tested in this PR
	- Flat hooks for `ft` (terminals may further shrink/expand for variants or ligatures)
- Potential further customization on specific arc-type combinations in `Bowl`
- Mask offsets, if needed
- Actual application to existing glyphs.


be5invis:
Any progress on this?

Logo121:
Sorry if I didn't make this clear but I don't really have much idea what to do with this and am looking for comments on the "Issues to be resolved".

It's technically "complete" but as I mentioned above, I'm not sure if using a namespace-returning function is the best approach here.

An example of applying these arc-hook handlers would be like this
https://github.com/be5invis/Iosevka/blob/ab30cad311a230b54555fae2644a94430eadee0a/packages/font-glyphs/src/letter/arc-hook.ptl#L336-L341
Which is pretty straightforward, but it would effectively construct a namespace 32 times in total if we consider all variant combinations, and likely several times of that when we apply to each `OBar`-using character, which *feels* wrong.

be5invis:
> Is using a namespace-returning function really appropriate here, considering it will be called a lot for different variants?

I've already using that so no worries :)

be5invis:
For letter-specific variants you can try not including these special details, and see if there's anytihng becoming strange.

`flex-params` is flexible enough to include fine details.

Logo121:
image

I've added a new test on a more compressed b (using `ь` as reference) which seems to show the problem if these two are skipped:
>  - `flatside` on the stem side of `OBar`. This is done for the standard `OBar`, but not the other variants.
>  - `yTerminal` for `OBar`.

But these are probably better handled by some rewritten `OBar` functions? (since it depends on the other hook/arc type)

I don't think my eyes are good enough to notice the difference `arch.(...)` makes though

be5invis:
I sort of want to drop `OBar` so perhaps you should add the "counterpart" knot generators. For example, for your `b`:

```
				include : dispiro
					ArcC1.arcStartL
					flatside.rd df.rightSB 0 XH SmallArchDepthA SmallArchDepthB 0
					ArcC2.arcEndL
```

We could probably change it to this

```
				include : dispiro
					ArcC1.arcStartL
					ArcC1.smoothToStraight
                                        ArcC2.straightToSmooth
					ArcC2.arcEndL
```

For `n` we could have

```
				include : dispiro
					ArcC1.arcStartL
					ArcC1.smoothToStraight
                                        Curl [ArcC1.smoothToStraight.x]
```

But this may introduce _more_ complexities (x-overshoot, etc.)

be5invis:
I think we need to cover the cases for `a`, `b`, `d`, `u`, `n`, `p`, `q`, `ь` (compressed bowl), for both dispiro and filling (use as mask).

Logo121:
Still working on it, current changes:
- Added the `ada`/`adb` compressing formula into `aesthetics.ptl` since a few functions use them already
- Added a whole `shoulder-bowl.ptl` for handling the OBar shapes and rewrote the existing "test shapes" to use these instead
- Added a new test which is basically a mask based on `a`'s shape. Still trying to fix the mask shapes. 


> We could probably change it to this
> 
> ```
> 				include : dispiro
> 					ArcC1.arcStartL
> 					ArcC1.smoothToStraight
>  					ArcC2.straightToSmooth
> 					ArcC2.arcEndL
> ```

The problem with this would be:
- In the compressed bowl case, the "smoothToStraight" and "straightToSmooth" becomes 1 knot, which means these 2 processes would likely require some interaction somehow
- On the `arcStart`/`arcEnd` side, their `ada`/`adb` values would still need to be compressed to prevent the problematic shapes shown above (crossing between the start and end part of the stroke)

So my idea was to still keep a OBar-like function (named `Bowl` in the code for now), but it takes 2 arc handler classes (any of `{Toothed Toothless Rounded Flat}`; whatever they end up being called) and handles the instantiation and archdepth compression within the namespace.

be5invis:
> In the compressed bowl case, the "smoothToStraight" and "straightToSmooth" becomes 1 knot, which means these 2 processes would likely require some interaction somehow

Perhaps do like this:

```
    ArcC1.arcStartL
    ArcC1.compressedBowl ArcC2
    ArcC2.arcEndL
```

Start/end `ada/adb` could probably calculated during creating things like `ArcC1 / ArcC2`. Maybe you can create a function that returns _both_ of them

```
     local {.upperHalf a .lowerHalf b} : MakeCompressedArcBuilders ......
```

Logo121:
> Maybe you can create a function that returns _both_ of them

Well that's essentially what I did in the `Bowl` function:
https://github.com/be5invis/Iosevka/blob/41177e953a671536c294d8087d6f1f471a583eef/packages/font-glyphs/src/letter/shared/shoulder-bowl.ptl#L27-L30

Since the `.compressedBowl` would depend on the 2 arcs anyway, I still think having the builder also act as a mediator would work better.

Logo121:
Mask and Shoulder shape should work as expected now
image

jmcwilliams403:
> I don't think my eyes are good enough to notice the difference `arch.(...)` makes though

`arch.`{`lhs`|`rhs`} is like calling `widths.`{`lhs`|`rhs`} again, which normally resets the stroke width back to `Stroke` (even if that wasn't the previously used value) _unless_ it has also been passed an explicit `(sw -- …)` argument inline, which then sets it to a _specific_ stroke width, as needed.

`centerAt` adjusts it left or right under italics/obliques using the aforementioned parameters and `TanSlope` (or something along those lines).

{`rtl`|`ltr`} gives it information on which direction it's going, and therefore which direction to adjust it (against the relative slope angle, with additional context provided by {`lhs`|`rhs`}).

Logo121:
@jmcwilliams403 By that I meant the difference between 
https://github.com/be5invis/Iosevka/blob/b527da404217292de5f8ab3f772c28dc90c73374/packages/font-glyphs/src/letter/shared.ptl#L418-L421
and simply doing
https://github.com/be5invis/Iosevka/blob/b03b3d5ff66dd1f979d3756f66d619b2c8eda66e/packages/font-glyphs/src/letter/shared/arc-hook.ptl#L104-L106

...for the "toothless" bowl/shoulder variants. On further inspection it seems like the main difference is the `alsoThru` knot, which is set to `DMBlend = 0.8` in practically all cases. The later `arch` part is actually just the `arch.{lhs|rhs}` broken down.

But that implementation (left) looks more-or-less indistinguishable from the simplified version used in the test that does not have the blend (right) 
image

jmcwilliams403:
> But that implementation (left) looks more-or-less indistinguishable from the simplified version used in the test that does not have the blend (right)

Could you show any of these images so far under italics? That's where `ArchDepthA`/`ArchDepthB` actually differ from each other anyway.
Also fwiw `TanSlope` is typically `0` when upright and non-zero under italics, from what I've gathered.

Logo121:
`b`: image

The whole test: (`a`, `b`, `c`, `ɑ`, `e`, `ь`, vaguely `a`-shaped mask, `տ`)
image

be5invis:
Masks should contain the entire internal of the shoulder/bowl/couner.

Also, please leverage `DivFrame` objects and/or `Box` objects.

packages/font-glyphs/src/letter/shared/shoulder-bowl.ptl
```diff
@@ -0,0 +1,244 @@
...
+				local df : DivFrame 1
+				define YeriBarPos 0.55
+				local yArcTop : XH * YeriBarPos + df.mvs / 2
+
+				local Bowl : BowlT ac1 ac2 df.leftSB df.rightSB yArcTop 0 SmallArchDepthA SmallArchDepthB (stroke -- df.mvs)
+
+				include : Bowl.shape.arcR
+				include : VBar.l df.leftSB Bowl.ArcEnd.yLB XH df.mvs
+
+		# mask test based on double storey a shape
```
be5invis: Masks should cover the entire internal of the shoulder/bowl/couner.

be5invis: Also you need to specify the _offset_ of masks. In our implementation masks typically have a tiny offset to avoid handling of occurrent outlines.

be5invis:
I think it is good enough now, you can now remove the "test" glyphs.

Issue #2085, opened 2023-11-11
Serifed (or serif-like) inward hooks (at least for `a`)


Logo121:
- [x] The requested variant shape does not go too far away from Iosevka's design.
- [x] The requested variant does not conflict with any characters in Unicode that Iosevka currently supports.
- [x] At least two monospace/programming fonts, created by different designers, supported the requested variant. Provide images below.

------

Kind of a leftover request stemmed from #1489, but basically there are quite a number of serif fonts (especially slab-serif, though mostly non-monospace) that treats teardrops terminals or inward hooks as serifs
![image](https://github.com/be5invis/Iosevka/assets/21302803/40497a71-2a96-4704-9e94-091b67fa6d2a)
and sometimes even "hypercorrecting" them as outward serifs
![image](https://github.com/be5invis/Iosevka/assets/21302803/cd2fbd2f-6efb-4d86-b2b2-2eb763ccf59b)
though the latter is much rarer.

But this request is mostly motivated by the fresh and hot Monaspace font, which in its "slab serif" variant, seems to consistently have serifs for inner hooks (or even just consider inner hooks and serifs the "same shape")
![image](https://github.com/be5invis/Iosevka/assets/21302803/72a527d5-bc85-41c4-a0bb-f9cdf49220d8)
(in a more "normal" serif font, only `c` and `s` will be serifed)

I don't personally think that all of them are needed, some of them do seem common enough that they can reasonably be added as variants.

Going off the programmingfonts.org list again:
![image](https://github.com/be5invis/Iosevka/assets/21302803/c7070a59-bb2d-44f6-9935-7d2c4bc4862c)
~~It seems like that `r` is the most common character to have a sharp "cornered" bend on the hook, making it more like a serif. This particular variant appears across different styles.~~ `r` is already implemented.

This is followed by `f`, and then `aj23569` with a few fonts making all of them have serif-like.  
I suggest `a` over the rest because it seems to me that in slab-serif fonts, it is actually more common to see a serifed hook `a` than what's represented by that list. A quick glance through [some slab-serif font catalog](https://www.fontspring.com/tag/slab-serif/) will give you a good enough idea on that.

---
Here are some fonts from fontspring for further reference:
Myrma (`fjy`):
![image](https://github.com/be5invis/Iosevka/assets/21302803/f77c9259-b6e8-431d-92b1-43e33d69327e)
Colon Mono (`23`):
[...]

PR #3012, opened 2025-12-24
Flat Hook `5`

### LINKED ISSUE #1287: More open “5” and “c”

Logo121:
Partially resolves #1287.

`5Ƽƽ𝟝⁵₅` (Bottom 3 lines are before build)
screenshot

This PR tries to add flat-hook `5` variants, which is the easiest case out of the 3 mentioned in that issue. `ec` are much more involved and would probably be a draft PR for another time (especially since those changes kinda conflicted with #2972 or something else).

Changes made in Arc-Hook/Bowl-Shoulder/other code:
- Added `hook` override in Bowl/ShoulderT.
	- Ideally we might want a more comprehensive override system on these, something like `arcStartParams` `arcEndParams` (with appropriate defaults) that get passed down to the corresponding Arc constructors. I've no idea how to implement these (it's not as simple as `$-flex-params` i think)
	- (This is not used now)
- Minor modification on `Flat` arcs to (try to) handle potential breaks in crowded spaces + slope.
- Fixed some logic on `box.pad()`.

`5` Variants:
- Modified `5.ptl` to use Arc-Hook. Didn't use BowlT since it had too many special cases or problems.
- Renamed `middle` variant stage to `arcs`, with a new `flat-hook` variant, which has flat middle + hook.
	- Theoretically `arched` middle + `flat` hook is possible but it doesn't look right (Pasted image 20251208045732).
	- Other possible middle/hook arcs:
		- Toothed/Toothless middles (by extension): Probably will look even worse than arched middle.
		- Flat-serifed, Toothless-serifed, Toothed hooks: The kind of serifed-hooks I mentioned in #2085. Probably better to handle them if/when we do serifed digits.
		- Toothless-nonserifed: _This would actually provide the "humanist open hooks" requested in #1287,_ if that's allowed.
- Some minor change in how the flat middle variant is handled. This is required to fix some broken shapes in max slant heavy.

### DIFF
--- a/changes/34.0.0.md
+++ b/changes/34.0.0.md
@@ -2,6 +2,7 @@
 * Add characters:
   - UNCERTAINTY SIGN (`U+2BD1`) (#2976).
 * Add `straight-vertical-sides-almost-flat-top` and `rounded-vertical-sides-almost-flat-top` variants for `W` and `w`.
+* Add `flat-hook` variants for `5`.
 * Refine shape of the following characters:
   - LATIN CAPITAL LETTER GHA (`U+01A2`).
   - LATIN SMALL LETTER GHA (`U+01A3`).
--- a/packages/font-glyphs/src/letter/shared/arc-hook.ptl
+++ b/packages/font-glyphs/src/letter/shared/arc-hook.ptl
@@ -193,7 +193,7 @@ glyph-block Letter-Arc-Hook : begin
 		export : define [hookEndR] : begin
 			return : list
 				hookend box.b (sw -- sw)
-				g4 box.r (box.b + hook) [heading Upward]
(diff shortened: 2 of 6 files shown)


22ljz:
@Logo121 Hello, could you help to make flat-hook for `S`?



It's so hard to understand PatEL. 😢

Issue #2878, opened 2025-09-27
Change the italic form of Aile


TimurGerasimov250:
Please make it true italics, like in sans. it would be more natural and flowing form. And also it would match Etoile to be both actual companions. with more cursive italics.

be5invis:
Again, Aile's target is to mimick neo-grotesque styles.

be5invis:
You can solve your problem by creating a [custom build](https://github.com/be5invis/Iosevka/blob/main/doc/custom-build.md). You can build a own variant that matches your need, or even distribute it as this font is licensed under OFF.

PR #3013, opened 2025-12-25
Add true italics for Iosevka Aile

TimurGerasimov250:
Iosevka Aile currently uses oblique-style italics, while the main monospaced families and Iosevka Etoile both provide true cursive italics.

For better style consistency across the default families, it would be great if Aile also had true italics similar in spirit to Sans, Slab, and Etoile:
- More natural, flowing cursive forms instead of simple slanted uprights.
- Visual consistency when mixing Sans, Slab, Aile, and Etoile in the same document or UI.
- A more coherent “companion” relationship between Aile and Etoile.

This would effectively make Aile a closer companion to Sans, Slab, and Etoile, with all three offering matching true italics rather than a mix of true italic and oblique styles.

This request is a follow-up to issue #2878 (please see that issue for related discussion and context) (Already closed)


be5invis:
> Iosevka Aile currently uses oblique-style italics

This is by design. Aile was created to become like neo-grotesque which usually have italics look a lot like oblique.

be5invis:
Your PR has zero change. Please close either this or #3014.

TimurGerasimov250:
B

Issue #3020, opened 2025-12-27
Complete Z Notation by adding 2 more symbols


TimurGerasimov250:
Hi Be5invis,
There are two very simple math symbols that are needed to complete: 

U+2A20 Z NOTATION SCHEMA PIPING, by copying code from U+226B MUCH GREATER-THAN

U+2A21 Z NOTATION SCHEMA PROJECTION, by copying code from U+21BE UPWARDS HARPOON WITH BARB RIGHTWARDS

Thanks, we can just quickly add these to the font to v34 and close the issue

TimurGerasimov250:
Add them to backlog to be added soon

PR #3021, opened 2025-12-27
Complete Z notation signs

be5invis:
#3020
  - Z NOTATION SCHEMA PIPING (`U+2A20`).
  - Z NOTATION SCHEMA PROJECTION (`U+2A21`).

### DIFF
--- a/changes/34.0.0.md
+++ b/changes/34.0.0.md
@@ -2,6 +2,8 @@
 * Iosevka private characters moved to U+F880..U+F8FF range.
 * Add characters:
   - UNCERTAINTY SIGN (`U+2BD1`) (#2976).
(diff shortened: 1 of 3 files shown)


dscorbett:
U+2A21 Z NOTATION SCHEMA PROJECTION should not be an alias of U+21BE UPWARDS HARPOON WITH BARB RIGHTWARDS. According to ISO/IEC 13568, page 24, “Schema projection ‘⨡’ and the mathematical toolkit character filter ‘↾’ are different Z characters”. In particular, U+2A21 is larger than U+21BE, similar to U+2A1F vs. U+2A3E or U+29F9 vs. U+005C.

PR #3065, opened 2026-01-22
Docker build: Avoid unnecessary repeated fetching of source tarballs and Node dependencies

why-trv:
Current Docker build workflow fetches and unpacks a source tarball and installs Node dependencies from scratch on **every** font build (`docker run`). This is just wasteful when iterating on a build plan, and I think the dependencies should rather be baked into the Docker image itself. 

It also doesn't make sense to fetch the source tarball at all since the user has presumably already cloned or downloaded the source. If they want to build an older version, they should checkout the appropriate tag instead, shouldn't they?

### DIFF
--- a/packages/font-glyphs/src/letter/latin/upper-n.ptl
+++ b/packages/font-glyphs/src/letter/latin/upper-n.ptl
@@ -24,81 +24,103 @@ glyph-block Letter-Latin-Upper-N : begin
 	define SLAB-CYRL-I  4
 	define SLAB-DIGAMMA 5
 
+	define pInkTrap 0.4
+
 	define [NShape] : with-params [bodyType slabType top left right [crowd 2] [crDiag 4]] : glyph-proc
 		local swDiag : AdviceStroke crDiag
 		local stroke : AdviceStroke crowd
-		local yEnd : match bodyType
-			[Just BODY-SYMMETRIC]    0
-			[Just BODY-ASYMMETRIC] : mix 0 top 0.375
 		local swEnd : match bodyType
 			[Just BODY-SYMMETRIC]    swDiag
-			[Just BODY-ASYMMETRIC]   stroke
+			__                       stroke
+		local swStart : match bodyType
+			[Just BODY-SYMMETRIC]    swDiag
+			__                       stroke
+		local xEnd   : right - swDiag
+		local xStart : left  + swDiag
+		local yEnd : match bodyType
+			[Just BODY-ASYMMETRIC] : mix 0 top 0.375
+			__                       0
+		local yStart : match bodyType
+			[Just BODY-ASYMMETRIC]   top
+			__                     : top - yEnd
 
 		include : union
 			match bodyType
 				[Just BODY-SYMMETRIC] : dispiro
 					flat left 0 [widths.rhs.heading stroke Upward]
-					curl left [mix 0 top 0.4] [heading Upward]
-					straight.up.end left top [widths.rhs.heading swEnd Upward]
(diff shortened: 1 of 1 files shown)


nobk:
@be5invis I would like to raise a serious concern regarding this merged PR. After reviewing the changes in depth, I believe this commit introduces **significant functional regressions** and a **problematic architectural shift** that tightly couples the toolchain with the font data, rather than keeping them independent.

1. Functional regression
With the new design, each Docker image can only build **one specific version** of the fonts. This means that **every time a new font version is released, all Docker users must rebuild their Docker images.** This is a major step backward in usability and maintainability.

2. Loss of toolchain–data separation
The previous design intentionally kept the toolchain independent from the font data. A single Docker image could build all versions, including future ones, without modification. This separation was a core advantage of the original approach.
The new implementation hard‑codes data into the image, creating unnecessary coupling and forcing users to rebuild images repeatedly. This is not a sustainable workflow.

3. The stated motivation does not hold up
The PR claims that the change is needed because:
- each build requires downloading data, and
- Node dependencies must be set up again,
- making repeated builds of the same version inconvenient.

However, this reasoning is flawed:
- The original design already supports repeated builds of the same version, and does so more flexibly.
- Users who need to rebuild the same version multiple times can already do so easily using the existing workflow.
- Not all users need repeated builds of the same version, and the new design imposes a heavy cost on everyone for a niche use case.

4. The original workflow already solves the stated problem
For example, the previous Docker image allows interactive multi‑step builds without rebuilding the image:
```
# Enter interactive mode
docker run -it --rm -v $PWD:/work --entrypoint bash fontcc
$ /build.sh contents::IosevkaSS06 ttf::IosevkaSS12
$ npm run build -- webfont::IosevkaSS15
$ cat myfont1.toml myfont2.toml myfont3.toml > private-build-plans.toml
$ npm run build -- ttf::myfont1 ttf::myfont2 contents::myfont3
$ exit
```
This workflow is both flexible and efficient, and it avoids the heavy coupling introduced by the new design.

### Request
Given the functional regressions and architectural issues introduced by this PR, I strongly recommend reverting this commit and restoring the previous Docker‑based build approach. The multi‑step build workflow shown above should be added to the documentation, as it already addresses the concerns raised in the PR without sacrificing flexibility or future compatibility.

why-trv:
@nobk 
Fair enough. I agree that baking in the particular version of source is perhaps too much.

However, there's a thing about the previous implementation that may need to be addressed: if one is working with a local clone or a fork, none of the changes to the internals get reflected in docker builds - it always uses the upstream source. So there's another kind of heavy coupling, isn't there?

nobk:
They can absolutely use Docker to build their own fork instead of the upstream repository. For example:
```
export GIT_FORK_REPO=$HOME/git/Iosevka_fork
# Enter interactive mode
docker run -it --rm -v $GIT_FORK_REPO:/work --entrypoint bash fontcc
$ npm install
$ npm run build -- ttf::userfont1
$ # anything else or debug
$ exit
```
So their local modifications can be built through Docker as well. The previous design already supports this workflow without forcing users to rebuild the entire image for every version.

That said, I believe that for developers—rather than end users—it is still important to set up a proper local development environment. It provides much finer control over the internals and makes iteration easier.

As for my own use case, I run an automated script that checks for new font releases daily. When a new version is available, it automatically pulls the latest source and builds with my customized .toml variants using Docker. This is a one‑shot automated process, not something I repeat manually.

why-trv:
OK, I get it. I guess you could say I should've RTFM on Docker, yet I believe this was underdocumented here in the repo. Going to prepare a new PR.

PR #3070, opened 2026-01-25
Refine rounded italic `e`

1024th:
The original entry stroke for the rounded italic `e` was designed to start from a slightly higher position and curve downwards to the right. This direction created a visually awkward transition, making the glyph appear unbalanced. The refined design now starts from the bottom left and smoothly curves upwards to the right.

The reversed versions of e are not changed because current design already aligns with handwriting conventions.

comparison


be5invis:
Looks good to me, it looks slightly more smooth. @1024th can you share the comparison at bold weights?

1024th:
Here is a comparison for light and bold weights.

be5invis:
@1024th Do you have images for heavy weight and normal width? (this will reveal potential broken geometry).

1024th:
ss15 IBM Plex Mono Style


ss03 Consolas Style


The `e` shape looks good to me, but the diagonal tail looks bad on normal width + heavy weight. Maybe I can fix it in a new PR.

be5invis:
LGTM

PR #3095, opened 2026-02-07
Docker: revert to fetching source at runtime

why-trv:
This partially reverts dc1b2913782d5e2907353387df69be78c72bf275, but also allows the source to be fetched from a fork repo, and documents the ways to use a local copy if desired. See discussion in #3065 for context

@nobk


nobk:
In `Interactive Mode` , user can also set environment variables to setup build version. Put `SOURCE=v34.1.0` before `/build.sh`.

```bash
docker run -it --rm -v $PWD:/work --entrypoint bash iosevka-builder
$ SOURCE=v34.1.0 /build.sh contents::IosevkaSS06 ttf::IosevkaSS12 webfont::IosevkaSS15
```

And I suggest /build.sh can run multiple times, so we need add ` rm -rf "$IOSEVKA_DIR"` before `mkdir -p "$IOSEVKA_DIR"` , do you think it's safe?

why-trv:
> And I suggest /build.sh can run multiple times, so we need add ` rm -rf "$IOSEVKA_DIR"` before `mkdir -p "$IOSEVKA_DIR"` , do you think it's safe?

Good point. Perhaps we can actually reuse it if `SOURCE` is the same as the previous run and delete otherwise. This will improve the ergonomics of repeated builds of the same thing (no need to switch between `/build.sh` and `npm run build` on initial and subsequent runs).

why-trv:
Went with caching based on URL+ETag in 56cc993a3f214c79e444e4ccb1e90573fd7feefe, not just `SOURCE` - to handle branch updates in between runs as well.

@nobk what do you think?

nobk:
I’m participating in this project purely as a user rather than a developer. The current functionality is already sufficient for my needs. After each new release, a single Docker command together with my custom configuration files produces the customized fonts I require, so this workflow is already well covered.

PR #3132, opened 2026-03-19
Add narrow hook variants for f, t, j, and r glyphs

jul-sh:
## Summary
- **f** (`narrow`): Curved hook with narrow crossbar and frame-relative positioning
- **t** (`narrowBentHook`): Bent hook with narrow crossbar
- **r** (`narrowHook`): Standard hook shape with narrower hook
- **j** (narrow): Bent hook with narrower descender hook

All three use `DivFrame para.advanceScaleF` for proper narrow cell width in quasi-proportional spacing. Glyph shapes use frame-relative coordinates (`df.leftSB`/`df.rightSB`) so they render correctly at both standard and narrow widths.

Scale factors for hooks and crossbars interpolate adaptively between conservative (monospace, `df.adws=1`) and expanded (quasi-proportional, `df.adws≈0.833`) values, so the variants look correct in both spacing modes.

## Motivation
In quasi-proportional settings, the default f, r, and t occupy full-width cells despite typically being rendered more narrowly in proportional fonts. These variants provide a middle ground between the standard shapes and the existing flat-hook/compact variants; keeping the familiar curved hook aesthetic while fitting a narrower cell.


mi2ebi:
what do they look like?

jul-sh:
image

be5invis:
Please show how they behaves at heavy weight.

jul-sh:
Good call to ask for heavy weight, see result below:

comparison

be5invis:
How about heavy italics and heavy oblique?

be5invis:
Also please try build a version with font style's width set to minimal (416) and maximal (720), show (thin and heavy) × (normal, italics, oblique).

be5invis:
From the current results the heavy doesn't look too good -- you may need to reduce the width of the hook's terminal. Luckily `[hookstart]` and `[hookend]` both support a `(swTerminal -- ...)` parameter to specify that.

jul-sh:
Updated! Here's the full matrix (thin + heavy) × (normal, italic, oblique) × (width 416, 500, 720):

comparison


And a heavy-weight close-up showing the hook terminals:

comparison_heavy


Good call on swTerminal; I'm now using mix sw [AdviceStroke 2.75 df.adws] [mix 0.35 0.1 pAdaptive] so the terminal taper scales with narrowness (stronger at w416, subtle at w720). Also raised the hook extension floor slightly for f and j so they have more room to breathe at extreme narrow widths.

PR #3133, opened 2026-03-20
Add dense ellipsis variant

jul-sh:
## Summary
- Adds a new ellipsis variant selector with dense option (smaller dots, tighter spacing)
- Cross-product with dot shape: round-normal, square-normal, round-dense, square-dense

The goal is a narrower ellipses for UI applications, where this is frequently used to truncate short text snippets in narrow fields; where horizontal space is scarce.

Screenshot 2026-03-20 at 12 50 19 AM


be5invis:
Do we need to distinguish dot's roundness vs ellipsis' roundness? If not, it should only have "dense" vs "loose" variants, and let dot-roundness feature to control the dot's shape.

You may need to utilize `[derive-multi-part-glyphs]`, which is our primary function to create such multi-selector variants.

In your case, your two "parts" are: one being the "selector" or "density" glyph, the other is the dot, and you need to assemble them in the code (for example the function parameter sent to `[derive-multi-part-glyphs]`).

jul-sh:
Updated the PR to separate ellipsis into two independent axes using nested select-variant:

  - Dot shape — reuses existing punctuationDot selector (round/square)
  - Density — new ellipsisDensity selector (normal/dense)

Leaf glyphs are created for each {dotShape} × {density} combination, then two levels of select-variant resolve the build plan settings; inner selects density (follows ellipsisDensity), outer selects dot shape (follows punctuationDot).

I ended up not using derive-multi-part-glyphs since both axes are build-plan-level selections rather than per-glyph CV alternates. Nested select-variant is the simpler fit here. Happy to rework if you'd prefer derive-multi-part-glyphs for the CV graph wiring though.

One thing to flag: the dense variant only produces a visible difference in quasi-proportional mode (width 667 vs 1000). In monospace the cell width is fixed, so the dots just shift slightly closer together — effectively invisible. Is that acceptable, or would you prefer the dense variant to also make the ellipsis narrower in mono (similar to how twoDotsLeader uses advanceScaleI)? Alternatively,we could skip the variant entirely and just make the QP ellipsis narrower by default, without a new selector.

comparison

PR #3138, opened 2026-03-22
Add `short-neck`{`2`} variants for Greek Lower Phi⁠/⁠Psi (`φ`, `ψ`), to match Lower T (`t`).

jmcwilliams403:
Basically, the ascending forms of these characters may have a slightly shorter neck than fully ascender-height or cap-height (or in `ss09`'s case, also lower than `t`-height), as is used by the fonts which these stylistic sets are based on.

The new variants are placed at the end via `nonBreakingVariantAdditionPriority = 20`.

For Phi's case, the variants are a sub-grouping of `straight`.
  * The existing `neo-hellenic` variant is also moved to come right after all three `straight-*` variants (after `nonBreakingVariantAdditionPriority = 20` is removed), but with its own `nonBreakingVariantAdditionPriority = 10` to keep it in the same spot as before (until the next breaking update).
  * Lower Phi Symbol (`ϕ`) follows all three `straight-*` variants, but not `neo-hellenic` or `cursive`.

Also add IPA localized forms for Greek Lower Psi (i.e. non-ascending) and Lower Lambda (i.e. serifed under slab) for legacy IPA and Americanist notation respectively.

```
```

`ss02`:
image
`ss06`:
image
`ss09`:
image
`ss13`:
image

### DIFF
--- a/packages/font-glyphs/src/letter/greek/psi.ptl
+++ b/packages/font-glyphs/src/letter/greek/psi.ptl
@@ -55,25 +55,24 @@ glyph-block Letter-Greek-Psi : begin
 		local fine2  : clamp fine Stroke coarse
 
 		local doFullSerifs : (df.width > para.refJut * 7) && (doSerifLT || doSerifRT) && doSerifMT && doSerifMB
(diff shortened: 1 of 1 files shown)


jmcwilliams403:
The build failure was apparently a connection timeout.

PR #3139, opened 2026-03-23
Refine diagonal tail parameters

1024th:
Make the depth and degree of diagonal tail adjustable parameters, mainly is to increase its length at width 600. The tails of `k` and `f` are adjusted.

## Help Wanted

**Current implementation fails at weight 900.** The smooth connections in `f` and `j` disappear. I have no idea why it fails. Removing the line

```text
alsoThruThem : FArc tailAngle 4alsoThruThem : FArc tailAngle 4
```

entirely from `DiagTail.F` in `packages/font-glyphs/src/letter/shared.ptl` seems to fix the problem, but it makes the connection less smooth. I would deeply appreciate some help or guidance.

图片
图片


be5invis:
Probably at 900 the inner side's turn just goes too tight so spiro lib cannot find out solutions? You can try increase or decrease the number in `FArc tailAngle 4` and see if there are improvements. Or reduce tail width for super-heavy weights.

1024th:
I tried `FArc tailAngle n` with `n = 1,2,3,4,5,6,7,8,9,10,11,50,100` and only `n=1` works. I think `n=1` is equivalent to removing this line.

I found increasing inner radius works: changing `define r : rIn + 0.5 * fine` to `define r : 2 * rIn + 0.5 * fine` gives me the connecting curve. But the inner connecting curves are broken. I found they are already broken in the original version:

图片

I think the ideal solution is to directly use straight lines for inner connection, but I don't how to achieve this.

be5invis:
I think we may probably need to move the knot before arcvh slightly up more

```
		export : define [R cx cy depth sw] : begin
			define rInY : 1 * DefaultTightBendInnerRadius
			return : list
				curl cx (cy + [FineSw sw] + rInY) # this
				arcvh 8 DiagTailSuperness
				F (+1) cx cy rInY DefaultAngle depth sw

		export : define [L cx cy depth sw] : begin
			define rInY : 1 * DefaultTightBendInnerRadius
			return : list
				curl cx (cy + [FineSw sw] + rInY) # this
				arcvh 8 DiagTailSuperness
				F (-1) cx cy rInY DefaultAngle depth sw
```

Probably change `[FineSw sw]` to `sw` or `[mix sw [FineSw sw] 0.5]`, etc.

1024th:
Changing `[FineSw sw]` to `sw` or `[mix sw [FineSw sw] 0.5]` doesn't help.

I think maybe a proper way to do this is to set the knots along the inner side of the corner and drawing the stroke on the other side? I experimented this idea on `f` and it looks good (see the image below). But this requires changing every usage of the DiagTail.

图片
 
 A fragment of the changed code is
 
 ```text
 		export : define [F sign cx cy rIn angle depth sw] : begin
			define tailAngle : angle / 180 * Math.PI
			define {cosAngle sinAngle} : superEllipse tailAngle
			define {dCosAngle dSinAngle} : superEllipseDerivative tailAngle

			define fine : FineSw sw
			define dxTurn rIn
			define dxTailStart : dxTurn + sinAngle * rIn
			define dyTailStart : (rIn + fine) - cosAngle * rIn
			define dxDepth : (depth - 0.25 * sw) * (+dSinAngle)
			define dyDepth : (depth - 0.25 * sw) * (-dCosAngle)
			return : list
				curl (cx) (cy + fine + rIn)
				arcvh 8 DiagTailSuperness
				# alsoThruThem : FArc (- Math.PI + para.slopeAngle / 180 * Math.PI) (- Math.PI / 2) 10
				g4.[if (sign > 0) "right" "left"].mid
					cx + sign * dxTurn
					cy + O + fine
					# widths.center fine
				alsoThruThem : FArc tailAngle 4
				flat (cx + sign * dxTailStart)             (cy + O + dyTailStart)
				curl (cx + sign * (dxTailStart + dxDepth)) (cy + O + dyTailStart + dyDepth)

		export : define [R cx cy depth sw] : begin
			define rInY : 1 * DefaultTightBendInnerRadius
			return : F (+1) cx cy rInY DiagTailAngle depth sw

		export : define [L cx cy depth sw] : begin
			define rInY : 1 * DefaultTightBendInnerRadius
			return : F (-1) cx cy rInY DiagTailAngle depth sw
 ```
 
 The caller needs to change the way to invoke `DiagTail.L` or `DiagTail.R`. Since the `widths.lhs` or `width.rhs` depends on the direction of the tail, I actually think the `dispiro` and the first knot should be drawn inside `DiagTail`, instead of outside it.

 ```text
 	glyph-block-export SmallFDiagonalTail
	define [SmallFDiagonalTail df barLeft] : begin
		local xBarMiddle : barLeft + [HSwToV HalfStroke]
		return : dispiro
			widths.lhs
			flat barLeft TINY [heading Downward]
			DiagTail.L barLeft Descender ([DiagTail.StdDepth df Stroke] - 0.5 * Stroke) Stroke
 ```
 
 Do you think this refactor is reasonable?

1024th:
Sorry, I seem to have misunderstood the `dispiro` function. Is it true that a continuous stroke should be drawn using a single `dispiro` function? Is it possible to switch from `widths.center` to `widths.lhs`/`widths.rhs`?

be5invis:
> Is it possible to switch from widths.center to widths.lhs/widths.rhs?

Yes

1024th:
I implemented the inner-corner knots. The diagonal tail of `k` is also adjusted. Preview:

图片
图片

图片

图片

图片

PR #3153, opened 2026-04-11
Optimize `cursive` glyphs for Cyrillic Lower Ve (`в`).

jmcwilliams403:
Also code cleanup surrounding `letter/latin/upper-b.ptl`.

### Motivation and Context

Basically, the default cursive Cyrillic Lower Ve glyphs resembled the `open` variants for `B`, so this changes the glyph to be `closed`, like real handwriting, as well as most contemporary italic fonts I checked.

Additionally, the bar position of `cyrl/ve.cursiveTall` is moved to the height of `AsymmetricBBarPos` because this variant is based on a handwritten form that looks identical to (cursive) Latin Lower `b`, with an enlarged lower bowl. This also makes it match `б`/_`д`_ more.

Greek Lower Beta is _very slightly_ changed, but only superficially, since I tried to preserve the original look feel of it as much as I could. Therefore, this disunifies `cyrl/ve.cursiveTall` from `grek/beta.cursive`, which technically adds two glyphs.

This also adds `startAf` and `endAf` arguments to `flatside` because there was previously no way to add a `heading` argument to the `curl` knot. I checked every instance of `flatside` in the code that uses the `af` argument, and the only uses were `[widths.`{`lhs`|`rhs`}` …]`, so I made both `startAf` and `endAf` default to the contents of the normal `af` argument since it appears to not disrupt anything if that usage gets copied over to the `curl` knot.

### Influenced Characters

  - U+03B2: GREEK SMALL LETTER BETA
  - U+03D0: GREEK BETA SYMBOL
  - U+0432: CYRILLIC SMALL LETTER VE

### Glyph Quantity

2

### Samples

`АаБбВвГгДд ΑαΒβϐΓγΔδ∂`

Default:
image
`ss03`:
image
Compared to Consolas:
image

### DIFF
--- a/changes/34.4.0.md
+++ b/changes/34.4.0.md
@@ -1,2 +1,3 @@
 * Add `above-baseline` variants for Greek Lower Chi (`χ`).
 * Add `tall` variants for Cyrillic Lower Ze (`з`).
+* Optimize glyphs for `cursive` and `cursive-tall` variants for Cyrillic Lower Ve (`в`).
--- a/packages/font-glyphs/src/common/shapes.ptl
+++ b/packages/font-glyphs/src/common/shapes.ptl
@@ -547,13 +547,17 @@ glyph-block CommonShapes : begin
 			local-parameter : ada -- ArchDepthA
 			local-parameter : adb -- ArchDepthB
 			local-parameter : ox  -- OX
-			local-parameter : af  -- nothing # Additional function for the control knots
+			# Additional function(s) for the control knots
+			local-parameter : af      -- nothing
+			local-parameter : startAf -- af
+			local-parameter : endAf   -- af
 
 			return : if (u - d > ada + adb)
 				list
-					flat (x + ox) (u - ada) af
-					curl (x + ox) (d + adb)
-				g4 (x + ox) [YSmoothMidL u d ada adb] af
+					flat (x + ox) (u - ada) startAf
+					curl (x + ox) (d + adb) endAf
+				list
+					g4   (x + ox) [YSmoothMidL u d ada adb] af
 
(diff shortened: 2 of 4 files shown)


be5invis:
ss03 Heavy's top loop is a little too tight.
image

jmcwilliams403:
> ss03 Heavy's top loop is a little too tight.

How about `[AdviceStroke2 2.0 3.5 Ascender]`:
image

If this is still too tight I can try `[AdviceStroke2 2 4 Ascender]`.

jmcwilliams403:
Here's `[AdviceStroke2 2 4 Ascender]`:
image

jmcwilliams403:
Here's after using `adaTop` and `adbBot` for the left-side arch depths:
image

be5invis:
You may need to change `local bowl : mix stroke top barPos` to something like `local bowl : [mix stroke top barPos] - 0.25 * stroke`, etc. `barPos` parameter could pass in a different value.

jmcwilliams403:
Here's it with `0.25 * stroke` proportionately subtracted from its `barPos`:
image

jmcwilliams403:
Returning `stroke` to `AdviceStroke2 2 3 Ascender` and subtracting abstractly `QuarterStroke / Ascender`:
image

jmcwilliams403:
With `Arch.adjust-x.`{`top`|`bot`}:

Default:
image
`ss03`:
image

jmcwilliams403:
I've given up on `AsymmetricBBarPos` for now. Here's `ss03` with normal `BBarPos`:
image

be5invis:
LGTM

PR #3155, opened 2026-04-13
chore: improve performance of dev-daily-build workflow via caching

jaschiu:
### Motivation and Context

Upon seeing dc156e65c0d9c06dbe93cb46afcc79a9567b9a78, I noticed that it could be a couple seconds faster, so I took what I have had at https://github.com/jaschiu/ljs-iosevka/blob/master/.github/workflows/build.yml to add efficient dependency caching. And `fail-fast: false` makes a job for one font that fails early not cancel other in-progress ones

### DIFF
--- a/.github/workflows/dev-daily-build.yml
+++ b/.github/workflows/dev-daily-build.yml
@@ -1,0 +1,52 @@
+name: "Dev Daily Build"
+on:
+  workflow_dispatch:
+    inputs:
+      reason:
+        description: "Reason for running manually"
+        required: false
+        default: "Manual check"
+  schedule:
+    - cron: "0 0 * * *"
+
+jobs:
+  build:
+    name: Build ${{ matrix.fontName }} font
+    if: github.repository == 'be5invis/Iosevka'
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        fontName: ["Iosevka", "IosevkaSlab", "IosevkaAile", "IosevkaEtoile"]
+
+    steps:
+      # Checkout repository into `iosevka` sub directory
+      - uses: actions/checkout@v6
+        with:
+          path: iosevka
+          ref: dev
+
+      - name: Use Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: lts/*
+          cache: npm
+
+      - name: Install ttfautohint
+        uses:  awalsh128/cache-apt-pkgs-action@v1
+        with:
+          packages: ttfautohint
+
+      - name: Build Font ${{ matrix.fontName }}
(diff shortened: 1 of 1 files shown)


.github/workflows/dev-daily-build.yml
```diff
@@ -16,30 +16,28 @@ jobs:
...
     strategy:
+      fail-fast: false
       matrix:
         fontName: ["Iosevka", "IosevkaSlab", "IosevkaAile", "IosevkaEtoile"]
 
     steps:
-      # Checkout repository into `iosevka` sub directory
       - uses: actions/checkout@v6
         with:
-          path: iosevka
```
be5invis, line 26: Why remove this? It doesn't slow things down

.github/workflows/dev-daily-build.yml
```diff
@@ -16,30 +16,28 @@ jobs:
...
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y --no-install-recommends ttfautohint
+        uses:  awalsh128/cache-apt-pkgs-action@v1
+        with:
+          packages: ttfautohint
 
       - name: Build Font ${{ matrix.fontName }}
         shell: bash
-        working-directory: iosevka
```
be5invis, line 42: Same as above

.github/workflows/dev-daily-build.yml
```diff
@@ -48,4 +46,4 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: ttc-${{ matrix.fontName }}
-          path: iosevka/dist/.ttc/${{ matrix.fontName }}/
```
be5invis, line 51: Same as above

be5invis:
One Iosevka build will take over 2 hours and the majority of time is inside its build process. Is it really worthy to save several seconds?

jaschiu:
> One Iosevka build will take over 2 hours and the majority of time is inside its build process. Is it really worthy to save several seconds?

Eh, there could be a rare chance that the ubuntu package index or npm registry has an outage, so the caching can bypass that. Feel free to close though

be5invis:
OK fine, then please resolve the conflicts and keep the `iosevka/` sub directory intact (aligning with PR actions).

jaschiu:
> OK fine, then please resolve the conflicts and keep the `iosevka/` sub directory intact (aligning with PR actions).

ok I restored `iosevka/`, seems that the workflow doesn't exist in dev yet and only on main but fixing now (will be an added file instead of modified unfortunately)

be5invis:
It is there now. And stop force-pushing things, even for PRs. I always use squashed merging.

be5invis (approved):

PR #3157, opened 2026-04-14
Add `cursive-interrupted`{`-tall`} variants for Cyrillic Lower Ve (`в`), equivalent to shape before #3153 .

jmcwilliams403:
### Motivation and Context

This allows people to select the existing shape of Cyrillic Lower Ve (`в`) used in Iosevka Version 34.3.0 and prior. It also provides parity with `standard-interrupted-`{`…`}.

If necessary, `cursive-interrupted` can also be made default for italic; However, this PR doesn't currently change that.

Variants are added to the end of the stack via `nonBreakingVariantAdditionPriority = 100`.

### Influenced Characters

* U+0432: CYRILLIC SMALL LETTER VE

### Glyph Quantity

2

### Samples

Before (Version 34.3.0):
image
After (`cyrl-ve = "cursive-interrupted"`):
image


jmcwilliams403:
The fixed build somehow failed in the hinting stage.

PR #3172, opened 2026-04-30
Add framed check and cross mark

### LINKED ISSUE #3171: Overrides ❌ emoji
Using 34.4.0 and Iosevka Term in my iterm2, the ❌ emoji is rendered as a small multiplication sign instead.

This feels like it was intentional but is there a way to disable that, or is this actually not the intended behaviour?

be5invis:
Fixes #3171

image
image


UaE5LeN23DMjGfr:
And... there goes the quick or intentional visual grab of **Symbol & Color** when those two are not used as a pair

There's already uncolored marks ✔️ & ✖️ which can mass replace the colored mark emoji if a user so chooses. This hijack is beyond scope

PR #3184, opened 2026-05-15
Unify `OBar`{`Left`|`Right`}`.`{`toothless`|`earless`}`Corner` to use `Arch.`{`lhs`|`rhs`}` `{`top`|`bot`}` (blend`{`Pre`|`Post`}` -- {})`.

jmcwilliams403:
This is for parity across the different terminal shapes in the font. Basically, it unifies `OBar`{`Left`|`Right`}`.`{`toothless`|`earless`}`Corner` with that of `nShoulder` and `uBowl` instead of using `DMBlend`. The changes here are completely invisible to my eyes.

Also make `DToothlessRise` the default `rise` argument instead of `SHook` since every instance of `OBar`{`Left`|`Right`}`.`{`toothless`|`earless`}`Corner` uses it.

Also make `g.doubleStoreyOpen` use `SHook`.

### DIFF
--- a/packages/font-glyphs/src/letter/cyrillic/che.ptl
+++ b/packages/font-glyphs/src/letter/cyrillic/che.ptl
@@ -37,13 +37,14 @@ glyph-block Letter-Cyrillic-Che : begin
 		local-parameter : adb       -- ArchDepthB
 		local-parameter : stroke    -- Stroke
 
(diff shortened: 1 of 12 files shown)


Logo121:
(is that basically what I mentioned in https://github.com/be5invis/Iosevka/pull/2957#issuecomment-3582055369?)

PR #3185, opened 2026-05-17
Add wave arrows pointing directly right/left

frangio:
### Motivation and Context

I [ran into](https://github.com/leanprover/cslib/blob/608cbe1b629a276abd3f2081f9b42dc766d8fd78/Cslib/Languages/LambdaCalculus/LocallyNameless/Stlc/Basic.lean#L46) the character ⤳ missing from Iosevka.

There is another similar but different character included, ↝ RIGHTWARDS WAVE ARROW (U+219D).

I defined a new glyph based on the existing wave arrow, but with a single wave, as well as the leftward version. I'm sure these can be improved.

### Influenced Characters

- U+2933: WAVE ARROW POINTING DIRECTLY RIGHT
- U+2B3F: WAVE ARROW POINTING DIRECTLY LEFT

### Glyph Quantity

2

### Samples

wave-arrows-normal-1

For reference, the existing wave arrow:

wave-arrows-normal-2

### Term Samples

wave-arrows-fixed

### DIFF
--- a/packages/font-glyphs/src/symbol/arrow/wave.ptl
+++ b/packages/font-glyphs/src/symbol/arrow/wave.ptl
@@ -28,13 +28,12 @@ glyph-block Symbol-Arrow-Wave : for-width-kinds WideWidth1
 	define o : O * 2
 
 	define waveArrowAmplitude : (Width - SB) * DesignParameters.arrow_size * (0.7 + 0.4 * MosaicWidthScalar)
+	define WaveSw : AdviceStroke (5.5 - MosaicWidthScalar)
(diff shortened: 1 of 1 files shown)


be5invis:
How do these glyphs look like under Term?

frangio:
Added Term samples in the PR description.

Logo121:
have you tried heavy or heavy+italic? it looks like it _might_ break

frangio:
I hadn't tried but it doesn't seem to break:

wave-arrows-all

frangio:
Ah, there was indeed an inconsistency in heavy mode compared to the preexisting wave arrows:

wave-arrows-heavy-compare

Stroke width is not set in the same way. I will open a PR.

PR #3186, opened 2026-05-17
Add letter variants with flat top and flat bottom.

### LINKED ISSUE #2519: [FR] Sans-serif single-story lower-case a vs o

jmcwilliams403:
### Motivation and Context

Resolves #2519 and #1269 .

This is a new type of {toothless|earless}-like terminal similar to #3169 ; This, like in that PR, is also implemented as an alternate "mode" of {`toothless`|`earless`}`Corner` that is accessed when the `rise` argument is `0` or undefined. It may also be serifed, unlike the original {`toothless`|`earless`}`Corner` shapes.

### Samples

```
```

Example build under **sans**:

Upright:
image
Italic:
image

Example build under **slab**:

Upright:
image
Italic:
image

### DIFF
--- a/changes/34.6.0.md
+++ b/changes/34.6.0.md
@@ -1,3 +1,5 @@
+* Add `flat-top` variants for `a`/`g`/`m`/`n`/`r`/`p`/`q`/`α`/`η`/`р`.
+* Add `flat-bottom` variants for `G`/`U`/`b`/`d`/`u`/`μ`/`µ`.
 * Refine shape of the following characters:
   - LATIN CAPITAL LIGATURE OE (`U+0152`).
   - LATIN LETTER SMALL CAPITAL OE (`U+0276`).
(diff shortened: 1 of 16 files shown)


be5invis:
I hope this is the last time we introduce a whole chunk of new glyphs purely for variants.

PR #3199, opened 2026-05-25
Fix #3198 .

### LINKED ISSUE #3198: Italic ӷ
Hello. Please fix the descender, it’s too long.

Image

jmcwilliams403:
This also affects `ʂ` for parity/symmetry.
It only applies to the two extra "hybrid-serifed" variants used exclusively by {`s`|`cyrl/ghe.italic`}`/descBase`.

`ӷ`

image


Martin-Costa129:
I think only the black italic would be changed.

PR #3278, opened 2026-08-25
Add instructions for installation on openSUSE Tumblewed

foursixnine:
### Motivation and Context

I've been using this font for years, also on openSUSE, and found that it was missing on the install instructions


foursixnine:
Hmmm this being a change in the readme, must also go to the dev branch?

